PR #3, opened 2016-09-01
One-liner to install latest bosh2 cli version

drnic:
Perhaps helpful to others to get latest greatest `bosh2` regularly whilst its being updated twice a day.


cppforlife:
@drnic no sha1 checking and also requires jq. i m thinking to potentially add `bosh sync` command which would always check with bosh.io to find latest checksum/url. wdyt?

drnic:
`bosh2 sync` would be great; also `apt-get install bosh2` for debian universes might be helpful too

PR #9, opened 2016-09-08
`spec.id` returns a unique identifier rather than `unknown` in bosh C…

cunnie:
…LI v2

- helpful for `bosh-cli`-deployed machines (no BOSH Director)
  on vSphere, esp. if > 2 Concourse workers who register
  with `spec.id` (if both are "unknown", the last one
  wins, the first one, forgotten)

[#129632439](https://www.pivotaltracker.com/story/show/129632439)

Signed-off-by: Dmitriy Kalinin <dkalinin@pivotal.io>


dpb587-pivotal:
Oops; pushed to develop instead.

PR #12, opened 2016-09-13
Statically compile bosh-cli linux binaries

ljfranklin:
- Fixes "no such file or directory" error on alpine docker images due to
  missing glibc
- CGO_ENABLED defaults to true if your build box is the same arch as the
  target arch


dpb587:
Hooray - I was about to send a PR for this, too! Thanks!

cppforlife:
Story: https://www.pivotaltracker.com/story/show/132192405

PR #15, opened 2016-09-27
Allow interpolated variable to contain `=`

ljfranklin:
- Use case - Interpolate public key `-v public_key=ssh-rsa ...G4/+VHa1aw==`


dpb587-pivotal:
Story: https://www.pivotaltracker.com/story/show/131323057

PR #29, opened 2016-10-20
If the release is an invalid URL, error includes URL

cunnie:
It would have saved me about 1 hour this morning if the error message had included the URL. Really. Don't ask.

Ran unit tests. Did not run integration or acceptance tests.

### DIFF
--- a/release/set/manifest/validator.go
+++ b/release/set/manifest/validator.go
@@ -43,7 +43,7 @@ func (v *validator) Validate(manifest Manifest) error {
 
 		matched, err := regexp.MatchString("^(file|http|https)://", release.URL)
 		if err != nil || !matched {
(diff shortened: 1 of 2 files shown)


dpb587-pivotal:
But... I want to ask. Off by one when reviewing the error message?

cppforlife:
@cunnie I m a bit concerned that it would print potentially some basic auth password, etc.

cunnie:
@cppforlife : Good point — we've gone to a lot of trouble to not leak passwords, and this would be a step backwards. Let me see if I can first redact any passwords from the URL before I print.

@dpb587 : I'm not sure what you mean by off-by-one, but what led up to this error was that I _specified the wrong manifest on the command-line_: e.g. `bosh create-env wrong-manifest.yml`

cunnie:
I'm closing this PR — if I have time to modify this PR to redact sensitive information, I'll open a new PR.

PR #40, opened 2016-11-07
Rename final_name -> name in config/final.yml

dpb587:
The field is already in a file named "final" and we no longer need to differentiate from a dev name.

Backwards incompatible.

Thoughts?

### DIFF
--- a/acceptance/assets/sample-release/config/final.yml
+++ b/acceptance/assets/sample-release/config/final.yml
@@ -1,2 +1,2 @@
 ---
-final_name: sample-release
+name: sample-release
--- a/releasedir/fs_config.go
+++ b/releasedir/fs_config.go
@@ -9,7 +9,7 @@ import (
 /*
 # final.yml
 ---
-final_name: cf
+name: cf
 min_cli_version: 1.5.0.pre.1001
 blobstore:
   provider: s3
@@ -29,7 +29,7 @@ type FSConfig struct {
 }
 
 type fsConfigPublicSchema struct {
-	FinalName string                   `yaml:"final_name"`
+	Name      string                   `yaml:"name"`
 	Blobstore fsConfigSchema_Blobstore `yaml:"blobstore,omitempty"`
 }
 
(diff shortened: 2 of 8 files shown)


voelzmo:
I like it. If we're not doing this now, we're doing this never :)
One suggestions: Can we create a script that migrates releases to work with this cli? So far, it would probably contain
* https://github.com/cloudfoundry/bosh-cli#yaml-encoding-issues
* removing possible symlinks in `blobs`
* now moving `final_name` to `name`
* ???

cppforlife:
i like it but it will break basic flow for older releases: git clone, bosh create-release, bosh upload-release.

i think we should make it backwards compatible by reading both names and falling back.

reset-release has been getting pretty close to becoming migration tool. it could be enhanced with rencoding and renaming.

voelzmo:
@cppforlife the flow for old releases is already broken, right? 
* If you have links in your `blobs` folder, it fails
  * you have to delete the folder and re-download blobs
* If you have binary format in the .yml files, it fails
  * you have to run the script linked in the readme
* If you have 'old-style' format for blobstore configuration in `config/final.yml` or `config/private.yml` it fails
  * you have to adapt those files manually
* Other existing things like `.dev_releases` or `config/dev_release.yml`(?) should also be removed as they're useless (not sure if harmful)

Sound like a pretty good case for a migration tool. WDYT?

dpb587:
Updated for backwards compatible, and rebased.

voelzmo:
Nor sure if I like backward-compatible logic in there. Shouldn't we error our and tell people to use a migration command first?

dpb587:
Since it was a fairly minimal change for backwards-compatibility, I thought
it was worth the improved UX when users clone older repos like Dmitriy
mentioned. A bit nicer for them to not have to worry and have local
modifications just to create a release, especially when they'll eventually
pull and then have some indecipherable merge conflicts later.

I still think a single migration command or reset-release which includes
this migration is a good idea.

Still not a fan?

On Mon, Nov 14, 2016 at 01:16 Marco Voelz <notifications@github.com> wrote:

Nor sure if I like backward-compatible logic in there. Shouldn't we error
our and tell people to use a migration command first?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/cloudfoundry/bosh-cli/pull/40#issuecomment-260284552>,
or mute the thread

cppforlife:
> the flow for old releases is already broken, right?

@voelzmo not quite. if new user comes in and clones the repo, it will work ok for them (only base64 will cause problems).

voelzmo:
@cppforlife you're right the complete happy-path still works even with the new cli. All other cases (as stated above) are broken.

PR #44, opened 2016-11-10
--out flag to pull out values out of manifest

cppforlife:
useful when trying to grab single configuration value and redirect it somewhere...

```
$ bosh bm ./manifests/example.yml --out /instance_groups/name=proxy/jobs/name=nginx/properties/ca_cert
# prints out found CA cert
```


dpb587-pivotal:
I like it.

What about naming it `--path` instead of `--out`?

Also, the dynamic type checking seems a bit odd for the multiline behavior. If a user starts relying on this command to extract a path (but the value wasn't a string for some reason) you're suddenly going to be receiving YAML when you were expecting a string. Did you consider a `--raw` option to make it more explicit? Or, possibly more unified, instead of a global `--json` a generic `--format`/`--output` option which defaults to `default`, and whose default for this command happens to be `yaml`?

PR #49, opened 2016-11-19
read CA cert only once so that it can be read from a pipe

cppforlife:
```
$ bosh alias-env my-bosh -e my-bosh.com --ca-cert <(bosh bm ~/workspace/my-bosh/manifest-creds.yml --path /director_ssl/ca)
```


cppforlife:
cc @dpb587 hopefully makes your bash script better.

tylerschultz:
master?

PR #50, opened 2016-11-21
Adds --state flag to create-env and delete-env commands

cdutra:
If not specified should still work as before.


cppforlife:
reviewed with @dpb587-pivotal

dpb587-pivotal:
and made a minor follow up commit 3c70706bd89f1694199efdb08a95e0662e4d6973

PR #61, opened 2016-12-02
Add subject key identifiers and additional key usages

jaresty:
Three changes to bring generated certs closer to those generated by
certstrap:
- Add subject key identifiers to certs and CAs
- Add authority key identifiers to CAs pointing to self
- Set maxpathlen and isCA flags

### DIFF
--- a/vendor/config_server/types/certificate_generator.go
+++ b/vendor/config_server/types/certificate_generator.go
@@ -1,10 +1,13 @@
 package types
 
 import (
+	"crypto"
 	"crypto/rand"
 	"crypto/rsa"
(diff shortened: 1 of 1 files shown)


jaresty:
[#152974306](https://www.pivotaltracker.com/story/show/152974306)

dpb587-pivotal:
Given this is vendored code and we don't recall who/what was originally needing this functionality, I'm going to close this. Feel free to resubmit to the upstream [config-server](https://github.com/cloudfoundry/config-server) repository if you recall the downstream project needs.

PR #69, opened 2016-12-09
Add support for scoped vars files

dpb587:
It would be helpful to allow scoping variables to avoid collisions. It allows `--vars-file` to have an `=` delimiter for this support, but certainly open to suggestions.

When a `aws.yml` vars file has...

    subnet_id: subnet-a1b2c3d4

And a manifest has...

    cloud_properties:
      subnet: ((iaas.subnet_id))

It could be interpolated with...

    bosh interpolate --vars-file iaas=aws.yml

For a result of...

    cloud_properties:
      subnet: subnet-a1b2c3d4

Thoughts?

### DIFF
--- a/director/template/vars_file_arg.go
+++ b/director/template/vars_file_arg.go
@@ -1,6 +1,8 @@
 package template
 
 import (
+	"strings"
+
 	bosherr "github.com/cloudfoundry/bosh-utils/errors"
 	boshsys "github.com/cloudfoundry/bosh-utils/system"
 	"gopkg.in/yaml.v2"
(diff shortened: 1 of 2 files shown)


director/template/vars_file_arg.go
```diff
@@ -29,6 +39,18 @@ func (a *VarsFileArg) UnmarshalFlag(filePath string) error {
 		return bosherr.WrapErrorf(err, "Deserializing variables file '%s'", filePath)
 	}
 
+	if fileScope != "" {
+		scopedVars := map[interface{}]interface{}{}
```
dpb587: This seems necessary due to [typing in `find_op`](https://github.com/cppforlife/go-patch/blob/5097ab27332566f622afb1549a5423ae0012b93f/patch/find_op.go#L86) which conflicts with [`StaticVariables`](https://github.com/cloudfoundry/bosh-cli/blob/5646ea8481b15e3db2dbeb2ec1ddc4bf2189569e/director/template/static_vars.go#L3). Otherwise this sort of error happens...

    Finding variable 'iaas.subnet_id':
      Expected to find a map at path '/subnet_id' but found 'map[string]interface {}'

Not sure if this is the correct solution.

cppforlife: it expects that all maps are `map[interface{}]interface{}`

cppforlife: Can also change StaticVariables to be `map[interface{}]interface{}` (https://play.golang.org/p/sVn3QLK3FX) and then nest them, instead of making a map.

dpb587: i'll give that another try; it did seem cleaner

dpb587:
Forgot about this. Will reopen if I get back to it later and figure it out.

PR #75, opened 2016-12-13
add git+*:// protocol support for upload-release command

cppforlife:
Found this useful to replace https sources when importing releases. @dpb587 also was leaning this way in terms of dealing with private releases.

Allows

```
$ bosh upload-release git+https://github.com/cppforlife/zookeeper-release
$ bosh upload-release git+https://github.com/cppforlife/zookeeper-release --version 0.0.1
```

and

```
releases:
- name: zookeeper
  version: 0.0.1
  url: git+https://github.com/cppforlife/zookeeper-release
```

Relies on `git` binary so this should also work with private repositories via `git+ssh://` and `git+file://`.

PR though is mostly about adding 2 integration tests for `create-release` and `upload-release` without running real director.


cmd/upload_release.go
```diff
@@ -34,52 +39,71 @@ func NewUploadReleaseCmd(
...
+		return bosherr.WrapErrorf(err, "Creating tmp dir for git cloning")
 	}
 
-	return c.uploadFile(opts)
-}
+	defer c.fs.RemoveAll(repoPath)
 
-func (c UploadReleaseCmd) uploadRemote(url string, opts UploadReleaseOpts) error {
-	version := semver.Version(opts.Version)
+	_, _, _, err = c.cmdRunner.RunCommand("git", "clone", opts.Args.URL.GitRepo(), "--depth", "1", repoPath)
```
dpb587: Might want to set `GIT_TERMINAL_PROMPT=0` to avoid infinite hangs if the user typos or tries to do an otherwise interactive git clone (like on a typo'd repo or one requiring authn).

dpb587:
Looks nice! Didn't realize `upload-release` had a `--name` argument. I suppose at some point it could figure out the release named once it cloned and double check whether the release exists at that point, too. Or maybe reuse ReleaseManager stuff and then it could support a `create` version, too.

I think private release support would need a bit more work; most will require `private.yml` for blob downloading.

Should remember to merge to `develop` instead of `master`.

cppforlife:
> I think private release support would need a bit more work; most will require private.yml for blob downloading.

right. some repos have read only creds -- not that that should be the recommended way. concourse can download the repo though and drop creds in there. as a side effect of this change:

```
releases:
- name: foo
  version: 123
  url: file:///tmp/build-whatever
```

works and picks up foo/foo-123.yml out of that repo.

PR #76, opened 2016-12-13
Fix ssh when there is no public key given by director

ArthurHlt:
If a bosh director doesn't give a public key when requesting bosh-cli respond with this error:

```
No RSA host key is known for 10.244.52.2 and you have requested strict checking.
Host key verification failed.
```

Indeed, the bosh-cli call this command to do the ssh:

`ssh -tt -o ServerAliveInterval=30 -o ForwardAgent=no -o PasswordAuthentication=no -o IdentitiesOnly=yes -o IdentityFile=/var/.../ssh-priv-key485211751 -o StrictHostKeyChecking=yes -o UserKnownHostsFile=/var/.../ssh-known-hosts485211751 10.244.52.2 -l bosh_3e32d12a96b04095 echo toto`

but when no public key is given the file `ssh-known-hosts485211751` is empty. 
It seems that the actual bosh-cli handle this error by removing the argument `-o UserKnownHostsFile=/var/.../ssh-known-hosts485211751` as we can see here: https://github.com/cloudfoundry/bosh/blob/master/bosh_cli/lib/cli/ssh_session.rb#L20-L25

This pull request do the same behavior as the current cli does, if the file is empty it removes this argument. This change fix the issue.

To recreate this issue, simply install a bosh-lite and just try to do a ssh on an deployment.

### DIFF
--- a/ssh/session.go
+++ b/ssh/session.go
@@ -11,14 +11,14 @@ import (
 )
 
 type SessionImpl struct {
-	connOpts ConnectionOpts
-	sessOpts SessionImplOpts
-	result   boshdir.SSHResult
+	connOpts       ConnectionOpts
+	sessOpts       SessionImplOpts
+	result         boshdir.SSHResult
 
 	privKeyFile    boshsys.File
 	knownHostsFile boshsys.File
 
-	fs boshsys.FileSystem
+	fs             boshsys.FileSystem
 }
 
 type SessionImplOpts struct {
@@ -26,10 +26,10 @@ type SessionImplOpts struct {
 }
 
 func NewSessionImpl(
-	connOpts ConnectionOpts,
-	sessOpts SessionImplOpts,
-	result boshdir.SSHResult,
-	fs boshsys.FileSystem,
+connOpts ConnectionOpts,
+sessOpts SessionImplOpts,
+result boshdir.SSHResult,
+fs boshsys.FileSystem,
 ) *SessionImpl {
 	return &SessionImpl{connOpts: connOpts, sessOpts: sessOpts, result: result, fs: fs}
 }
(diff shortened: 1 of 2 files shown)


ssh/session.go
```diff
@@ -62,9 +62,13 @@ func (r *SessionImpl) Start() ([]string, error) {
 		"-o", "IdentitiesOnly=yes",
 		"-o", "IdentityFile=" + r.privKeyFile.Name(),
 		"-o", "StrictHostKeyChecking=yes",
-		"-o", "UserKnownHostsFile=" + r.knownHostsFile.Name(),
 	}...)
-
+	statHostFile, _ := r.knownHostsFile.Stat()
```
ArthurHlt: I skip the error because at this step the file can only exists

ArthurHlt:
I'm sorry, my IDE format some lines, there is only few lines. I hope it will not be too annoying

cppforlife:
@ArthurHlt why does director not return host fingerprint?

ArthurHlt:
@cppforlife I have absolutely no ideas, I actually suppose that it's coming from bosh lite. and like the actual cli in ruby seems to handle this case, I've supposed that there was a reason. 

Troubleshoot the director it's not something that I'm familiar to do. 

What I could see for sure is that the response from director didn't give any public host key when performing a SetupSsh.

cppforlife:
@ArthurHlt let's take a look at few things:
- `bosh env` to see which director version
- `bosh deployments` to see which stemcell you are using
- `bosh task X --result` (whatever task id is shown when doing ssh) to see what agent responds with -- should include host fingerprint

ArthurHlt:
Thanks you for this, I've seen that my stemcell was an ubuntu lucid in version 64 ....

With the latest trusty stemcell all is good without this fix.
I didn't expect that it was the stemcell (more the vm) which was giving the public host key so I didn't look at the stemcell, that's actually make sense ...

I don't know if you want to support very old stemcell, for me there is no reason to merge this.

Thank you again, next time I will do an issue before troubleshoot the entire world.

cppforlife:
We'll close this out then. Let us know if you hit any other problems.

PR #78, opened 2016-12-19
Support Context Id in SDK

Samze:
CF BOSH Tracker story [#134702861](https://www.pivotaltracker.com/n/projects/956238/stories/134702861)

## Why

As a bosh user
I can assign a context to bosh deployments and errands
So that I can query a group of bosh tasks

## What

This PR extends the BOSH Go SDK to support using the new [`X-Bosh-Context-Id'](https://github.com/cloudfoundry/bosh/pull/1531). This change is not exposed in the CLI, only in the Go Library, so the UX is unchanged.

- Director interface extended with  a `WithContext(string)` function that returns a Director whose subsquent requests will include the `X-Bosh-Context-Id` header.
- Currently limited to GET/POST/PUT (see below)
- Header added at http client request level so no interface changes to operations.
- Director interface extended with a `FindTasksByContextId(string) ([]Task, error)`
- Task now has a `contextId` property

**Missing Delete support**
The Bosh-Utils HttpClient package only supports customising requests for Get/Post/Put but not Delete. [This PR](https://github.com/cloudfoundry/bosh-utils/pull/7) has been submitted to add the functionality to DELETE. 

However, since the vendored `18e54c` sha, the head of Bosh-Utils has breaking changes to other areas of the Bosh CLI, e.g. [Blobstore interface changes](https://github.com/cloudfoundry/bosh-utils/commit/ed39ea3a92b20becbfb16a2b02cf52e58aad5c26#diff-e76fb5507231062ad745b68e69863a3d). So even when that PR gets merged there will be additional work to integrate with head. @cppforlife / @avade thoughts?

Sam

### DIFF
--- a/director/client.go
+++ b/director/client.go
@@ -23,3 +23,12 @@ func NewClient(
 	taskClientRequest := NewTaskClientRequest(clientRequest, taskReporter, 500*time.Millisecond)
 	return Client{clientRequest, taskClientRequest}
 }
+
+func (c Client) WithContext(contextId string) Client {
+	clientRequest := c.clientRequest.WithContext(contextId)
+
+	taskClientRequest := c.taskClientRequest
+	taskClientRequest.clientRequest = clientRequest
+
+	return Client{clientRequest, taskClientRequest}
+}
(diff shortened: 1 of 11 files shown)


Samze:
Additionally, what do you think of changing `WithContext` to `WithContextID` as context is an overloaded term in Go and Bosh?

PR #87, opened 2017-01-03
add %i so that bosh scp can download files into instance namespaced directories

cppforlife:
was trying to download persistent data for all instances in a deployment. ended doing quick change so that `bosh scp :/var/vcap/store /tmp/inst-%i` makes scp create directory per instance and put content there.

cc @dpb587

### DIFF
--- a/ssh/scp_args.go
+++ b/ssh/scp_args.go
@@ -47,6 +47,11 @@ func (a SCPArgs) ForHost(host boshdir.Host) []string {
 			pieces[0] = fmt.Sprintf("%s@%s", host.Username, host.Host)
 		}
 
+		for i, _ := range pieces {
+			// Replace special identifier '%i' with specific instance
+			pieces[i] = strings.Replace(pieces[i], "%i", host.IndexOrID, -1)
+		}
+
 		args = append(args, strings.Join(pieces, ":"))
 	}
 
--- a/ssh/scp_args_test.go
+++ b/ssh/scp_args_test.go
@@ -14,7 +14,7 @@ var _ = Describe("SCPArgs", func() {
 	)
 
 	BeforeEach(func() {
-		host = boshdir.Host{Username: "user", Host: "127.0.0.1"}
+		host = boshdir.Host{Username: "user", Host: "127.0.0.1", IndexOrID: "id"}
 	})
 
 	Describe("AllOrInstanceGroupOrInstanceSlug", func() {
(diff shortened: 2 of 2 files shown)


dpb587:
I don't really like `%i` being a string (vs int), and the ambiguity between it representing "instance group name", "ID", or "index", all starting with "i". Target directory doesn't include instance group name which would be useful on more diverse deployments.

My vote would be for `xargs` to gain a bit more flexibility at the cost of conciseness...

    bosh instances | cut -f1 | xargs -n1 -I%s -- bosh scp -r %s:/var/vcap/store /tmp/%s/.

Would also want to have cli `mkdir -p` the target directories, probably conditional on if the path is suffixed with `/.`.

If `bosh instances` accepted a `-l`/`--list`/`--slug-only`-type thing, `cut` could be skipped. I'd use that option in other automation scripts, too.

One downside to `xargs` is it's much less efficient with the director. Could add `-P10` to xargs and bring back some parallelization though.

cppforlife:
> I don't really like `%i` being a string (vs int), and the ambiguity between it representing "instance group name", "ID", or "index", all starting with "i". 

fair enough. if we would go this route, any other naming suggestions for the id?

> Target directory doesn't include instance group name which would be useful on more diverse deployments.

ids are in theory are unique enough.

> My vote would be for `xargs` to gain a bit more flexibility at the cost of conciseness...

    bosh instances | cut -f1 | xargs -n1 -I%s -- bosh scp -r %s:/var/vcap/store /tmp/%s/.

that should also work but feels a lot heavier. does xargs allow to continue working even if one of the commands fails or does `|| true` have to be involved?

> Would also want to have cli `mkdir -p` the target directories, probably conditional on if the path is suffixed with `/.`.

scp seems to be makind a directory automtically. probably not nested ones though. not sure what `/.` is? just a slash? 

> If `bosh instances` accepted a `-l`/`--list`/`--slug-only`-type thing, `cut` could be skipped. I'd use that option in other automation scripts, too.

how about something like `--cols name,state` or something like that and done generically? or should we expand `--path` functionality?

> One downside to `xargs` is it's much less efficient with the director. Could add `-P10` to xargs and bring back some parallelization though.

possibly i guess. im still kinda liking %something but i thing xargs approach should be documented with all the bells and whistles.

cppforlife:
would be interesting to see bosh scp upload side of things as well. if i want to load bunch of data and i have directory with bunch of id subdirs...

dpb587:
If we're supporting something other than `xargs`, I'd prefer reusing our interpolation style, rather than fmt, although it would require paths to be single-quoted. Maybe it could interpolate from `instance_group` (something shorter?), `index`, `id`, `deployment`.

    bosh scp :/var/vcap/store `/tmp/inst-((id))`
    bosh scp :/var/vcap/store `/tmp/((deployment))-((instance_group))-((id))`

I agree, ids are unique enough to not need instance group. Was more for the case where you have a deployment with master/slave groups with slightly different directory structures where it might be helpful to keep track.

I agree, `xargs` does feel a bit heavyweight. As long as an invocation doesn't exit with 255, `xargs` will continue executing.

The `/.` suffix was just an idea so CLI could have something to detect if the target is going to be a directory (for the case of nested directories from slugs). Not important if we can do a single level.

I like the `--cols` idea.

scp upload could do the same interpolation; I could see that being useful.

I prefer `((id))` to `xargs` to `%i`.

dpb587-pivotal:
Merged via caa2b3f5dcbb2530d54c189cae8ac0f2de747f4f

PR #115, opened 2017-02-06
Show one event as a table flipped

barthy1:
[#136637059](https://www.pivotaltracker.com/story/show/136637059)

Signed-off-by: Konstantin Maksimov <kmaksimov@ru.ibm.com>

### DIFF
--- a/director/interfaces.go
+++ b/director/interfaces.go
@@ -24,6 +24,7 @@ type Director interface {
 	FindTasksByContextId(string) ([]Task, error)
 
 	Events(EventsFilter) ([]Event, error)
+	Event(int) (Event, error)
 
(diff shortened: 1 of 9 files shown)


barthy1:
uses https://github.com/cloudfoundry/bosh/pull/1577

director/interfaces.go
```diff
@@ -24,6 +24,7 @@ type Director interface {
 	FindTasksByContextId(string) ([]Task, error)
 
 	Events(EventsFilter) ([]Event, error)
+	Event(int) (Event, error)
```
cppforlife: event id should be string just like in the EventsFilter

PR #122, opened 2017-02-09
update-dep script enhancement

allomov:
Hey, all.

Right now, `./update-dep` script can fail if you fetch a package with circular symbolic links. For instance, if you'll run `./update-dep` with `github.com/cloudfoundry/bosh-cli` you'll get the following error:

```
 → ./update-dep github.com/cloudfoundry/bosh-cli
+ '[' 1 -ne 1 ']'
+ package_name=github.com/cloudfoundry/bosh-cli
+ go get -d github.com/cloudfoundry/bosh-cli
++ echo github.com/cloudfoundry/bosh-cli
++ sed 's/\.\.\.$//'
+ package_path=github.com/cloudfoundry/bosh-cli
+ rm -rf vendor/github.com/cloudfoundry/bosh-cli
+ mkdir -p vendor/github.com/cloudfoundry/bosh-cli
+ cp -r /Users/allomov/work/altoros/vmware/cf-audit/go/src/github.com/cloudfoundry/bosh-cli/ vendor/github.com/cloudfoundry/bosh-cli
cp: /Users/allomov/work/altoros/vmware/cf-audit/go/src/github.com/cloudfoundry/bosh-cli//vendor/github.com/bmatcuk/doublestar/test/abc/working-symlink: Too many levels of symbolic links
cp: /Users/allomov/work/altoros/vmware/cf-audit/go/src/github.com/cloudfoundry/bosh-cli//vendor/github.com/bmatcuk/doublestar/test/broken-symlink: No such file or directory
```
To fix it you just need to change `cp` opts from `-r` to `-PR` not to follow the links.

The script can be used by other projects or people, so I've decided to place fix here.

Both Mac OS and Linux support `cp -RP` options.

### DIFF
--- a/update-dep
+++ b/update-dep
@@ -17,7 +17,7 @@ package_path=$(echo ${package_name} | sed 's/\.\.\.$//')
 rm -rf vendor/${package_path}
 mkdir -p vendor/${package_path}
 
-cp -r $GOPATH/src/${package_path}/ vendor/${package_path}
+cp -RP $GOPATH/src/${package_path}/ vendor/${package_path}
 
 pushd $GOPATH/src/${package_path}
 set +e


allomov:
I also think that we can do better in [this place](https://github.com/cloudfoundry/bosh-cli/blob/1468871d2cb340c3eed98bd73576abb6e9aab647/update-dep#L54). We can not just remove the internal vendor folder, but move them to the vendor folder of the current package. This will also see changes in dependent packages in git.

The only thing is that the dependent packages are not be stored into `deps.txt`, but that's okey since we have the reference to the package we are adding right now.

We can replace this line to something like this:

```
vendor_folders=$(find . -name vendor -type d -mindepth 2 | sort -r)

# This may fail because of tests assets like symlinks and files without right permissions
set +e
while read -r vendor_folder; do
    echo "copying contents of $vendor_folder to vendor"
    cp -RP $vendor_folders/* /tmp/v1
    echo "removing $vendor_folder"
    rm -rf $vendor_folders
done <<< "$vendor_folders"
set -e

```

What do you think?

tylerschultz:
Hi @allomov, Thanks for your PR.

We have reservations about your second proposal. The change you propose would overwrite top level dependencies to versions that we may not want. We like the idea of being deliberate as to what version of the dependency we pick up. WDYT?

allomov:
@tylerschultz thank you for merging this changes.

At the moment sub-vendored dependencies are removed to avoid having different types in vendored and their sub-vendored packages. This approach makes sense, but it adds complexity in maintaining such dependencies. You need to ensure that interfaces, types and methods are not changed across different versions of dependencies. For me, if you copy subvendored dependencies to a root level of `vendor` folder you are able to see difference using git status, also you'll be able to check this changes by running tests. But it really depends on your workflow.

Could you please tell why you decided to use script driven dependencies update and not using existing package managers?

tylerschultz:
HI @allomov,
We switched to using our own script when we were feeling acute pain from community vendor tooling. We wrote our own scripts to enable us to move forward. We don't have a good reason anymore. We're long overdue to revisit using tooling now that things are settling out. We hope to do some of this work in the near future.

allomov:
I hope golang will come with one solution in closest releases.

[dep](https://github.com/golang/dep) tool is developed under `golang` org and looks promising, still they have warning that its interface may change.

PR #135, opened 2017-02-15
Use ui writer for progress bar output

drich10:
- This should not be merged until https://www.pivotaltracker.com/story/show/139968731
- This fix only fully applies to TrackUpload, TrackDownload will be missing a
  newline since it does not use the ReadCloserProxy. The rest of the
  progess bar will still be printed through the UI object.

Signed-off-by: Dan Wendorf <dwendorf@pivotal.io>

### DIFF
--- a/ui/file_reporter_test.go
+++ b/ui/file_reporter_test.go
@@ -8,16 +8,17 @@ import (
 	. "github.com/onsi/gomega"
 )
 
-var seekCalls []interface{}
+type CallTracker struct {
+	Seeks  []interface{}
+	Closes int
+}
 
-type FakeSeekableReader struct{}
+type FakeSeekableReader struct{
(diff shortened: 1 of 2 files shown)


ui/file_reporter_test.go
```diff
@@ -9,6 +9,7 @@ import (
 )
 
 var seekCalls []interface{}
+var closeCalls int
```
cppforlife: let's move these variables into FakeSeekableReader

cppforlife:
note: this PR depends on fixing RequestRetryable.

ui/file_reporter_test.go
drich10: Updated

cppforlife:
@wendorf done, and merged.

Issue #133, opened 2017-02-14
bosh cli should remove url when version:create


shalako:
When `version:create` is present in a manifest and `url` contains a file path, when running `bosh deploy` cli will generate a dev release from the release file path, and replace `version:create` with the dev release version.

Starting with
```
- name: routing
  url: file:///tmp/build/d5affdf1/release
  version: create
```

On deploy the CLI replaces `version:create` with a version
```
- name: routing
  url: file:///tmp/build/d5affdf1/release
  version: 0.145.0+dev.1487031596
```

The deploy succeed if the client has the dev release in this path, as would the client running the deploy the first time when `version:create`. However once the deployment succeeds, if another client downloads this manifest the url is meaningless and a subsequent deploy will fail.

As we believe the bosh cli is already replacing `version:create` with the dev release version, we believe it should also remove the `url` property.

@dsabeti

cppforlife:
after thinking about it, i think it makes sense to remove url when `version: create` is used before sending manifest to the director.

dsabeti:
@cppforlife: https://github.com/cloudfoundry/bosh-cli/pull/136

cppforlife:
Fixed in 2.0.3

shalako:
thank you, @dsabeti

PR #136, opened 2017-02-15
Remove file url from manifest when using 'version: create'

dsabeti:
See this issue: https://github.com/cloudfoundry/bosh-cli/issues/133


cppforlife:
@dsabeti nice, thx.

PR #138, opened 2017-02-17
Cross platform paths

Fydon:
Expected paths are Linux specific, so instead use filepath to return the path for the current OS.

I didn't change [this test][1], as the code would need to be the following, which I'm not sure if it is correct:

    Expect(result).To(Equal(filepath.Join("/fake/manifest/path", "fake", "relative", "path", "file.tgz")))

Should [this][2] be using [`path.Join`][3] instead of [`filepath.Join`][4], so that `/` is used on Windows builds as well?

I could make that change if that is fine.

[1]: https://github.com/Fydon/bosh-cli/blob/feature/CrossPlatformPathsInTests/common/util/file_helper_test.go#L36
[2]: https://github.com/Fydon/bosh-cli/blob/feature/CrossPlatformPathsInTests/common/util/file_helper.go#L32
[3]: https://golang.org/pkg/path/#Join
[4]: https://golang.org/pkg/path/filepath/#Join


Fydon:
Is the [fake file system][1] a unix like file system or should it behave like the file system that it is being used to test?

[1]: https://github.com/cloudfoundry/bosh-utils/blob/master/system/fakes/fake_file_system.go

tylerschultz:
@Fydon you ask great questions.

@monkeyherder and I are trying to reason through the implications.

To your first question, we believe the test is correct in it's current form. If the code is going to be able to read the file, it will need to be path'd correctly for the given platform. It is also interesting to understand if the paths in the manifests with unix style separators are expected to work on windows platforms. This feels like a bit of a product question. cc @cppforlife 

To your second question: We believe the fake file system should behave like the os filesystem would behave for a given platform. If the os filesystem needs updating, the fake filesystem should be updated to simulate that behavior. It seems the fake file system should base it's behavior based on the platform it's running on. (We intend to run the test suites on windows). It seems we don't have a solid answer here and we'll have to figure it out as we learn more.

Fydon:
@tylerschultz Thank you. Unix style paths do work on Windows, which makes this more complicated, as a test could pass for a given piece of code and yet not work when use in a specific case. For example the first change of this pull request just allowed some of the tests to pass, but the second change of the pull request enabled the use of `doublestar`'s Glob function (when it is update #139) which uses the [OS specific path separator][1].

I'll create a pull request about the proposed `bosh-utils` path changes.

Thank you. It would be great to have Windows tests running for `bosh-cli`.

This pull request makes a lot of the paths be OS specific, but there are still some tests that need to be updated to ensure that the code will work if given Windows paths, e.g. [fs_config_test.go][2]. There are also still some tests that are failing.

[1]: https://github.com/bmatcuk/doublestar/blob/master/doublestar.go#L166
[2]: https://github.com/cloudfoundry/bosh-cli/blob/master/releasedir/fs_config_test.go

Fydon:
@tylerschultz I forgot to mention that on Windows if you use `/file` as a file location for a command then you get `unknown flag /file`. This is why I needed to make changes like [this][1] for `factory_test.go`.

[1]: https://github.com/Fydon/bosh-cli/blob/84b4d148d63b7e051af398934fce07455aae04f2/cmd/factory_test.go#L45

PR #143, opened 2017-02-21
implement `repack-stemcell` command

cunnie:
`bosh repack-stemcell` allows the reconfiguration of an existing stemcell, allowing changes to the following:

* name
* version
* cloud-properties

https://www.pivotaltracker.com/story/show/139041167

Typical usage:
```
bosh repack-stemcell \
  --name bosh-aws-xen-hvm-ubuntu-trusty-go_agent \
  --version 3333.3 \
  --cloud-properties "ami: {}" \
  ~/Downloads/light-bosh-stemcell-3363-aws-xen-hvm-ubuntu-trusty-go_agent.tgz \
  /tmp/custom-stemcell.tgz
```

happy Tuesday

💟
Brian and Zak


zaksoup:
@cppforlife we chose not to remove the `GetExtractedPath` method from the interface for an `ExtractedStemcell`. Besides the name *Extracted*Stemcell implying that there will always be a local location of extraction this method was added to [remove `ImagePath` from the Manifest struct](https://github.com/cloudfoundry/bosh-cli/commit/712483ca2d80c539060ef1cd883fe653f734a561#diff-3f4d08aec9f1412e576882799b86f005L86) as `ImagePath` is not actually a property of the stemcell.MF and didn't make sense in our in-memory representation of the manifest.

cunnie:
@cppforlife ping

PR #155, opened 2017-02-28
Support custom VM metadata for bosh-init manifests

dpb587:
Similar to https://github.com/cloudfoundry/bosh/issues/936, but for bosh-init manifests. Note, that manifest tags can intentionally override init-generated ones (which is a bit [different](https://github.com/cloudfoundry/bosh/blob/c6c419f9eff844c35c8c7307e977427f05a23a51/src/bosh-director/lib/bosh/director/vm_metadata_updater.rb) than director behavior).

```
name: bosh
tags:
  director: *director
properties:
  director:
    name: &director my-prod-bosh
```

Thoughts?

### DIFF
--- a/cloud/cloud.go
+++ b/cloud/cloud.go
@@ -35,12 +35,7 @@ type cloud struct {
 	logTag       string
 }
 
-type VMMetadata struct {
-	Director   string `json:"director"`
-	Deployment string `json:"deployment"`
-	Job        string `json:"job"`
-	Index      string `json:"index"`
-}
+type VMMetadata map[string]string
 
 func NewCloud(
 	cpiCmdRunner CPICmdRunner,
--- a/cloud/cloud_test.go
+++ b/cloud/cloud_test.go
@@ -296,10 +296,10 @@ var _ = Describe("Cloud", func() {
 		It("calls the set_vm_metadata CPI method", func() {
 			vmCID := "fake-vm-cid"
 			metadata := VMMetadata{
-				Director:   "bosh-init",
-				Deployment: "some-deployment",
-				Job:        "some-job",
-				Index:      "0",
+				"director":   "bosh-init",
+				"deployment": "some-deployment",
+				"job":        "some-job",
+				"index":      "0",
 			}
 			err := cloud.SetVMMetadata(vmCID, metadata)
 			Expect(err).ToNot(HaveOccurred())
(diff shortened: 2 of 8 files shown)


cppforlife:
@dpb587 how do you feel about pulling out tags into runtime config and have create-env take in `--runtime-config` flag? eventually i'd like us to also accept `--cloud-config` and `--cpi-config`.

dpb587:
> for now runtime config can only have tags.

not sure what that means. cli doesn't do runtime-config yet, and v2 deployment manifests [support](http://bosh.io/docs/manifest-v2.html#tags) tags. or are you suggesting that's no longer supported?

i don't have an issue with `--runtime-config` supporting tags, but, assuming older-style manifests will still be supported by bosh-init, i'd rather not have to create a separate runtime-config file just to have the `tags` field (especially if/since regular deployment manifests can have tags already)

cppforlife:
> v2 deployment manifests support tags

silly me. yeah then it makes sense.

but yeah i guess at some point i see us bumping bosh-cli to be able to deal with separate config files. we'll pull this one in meanwhile.

tylerschultz:
Are we concerned that it is possible to overwrite the vm metadata bosh 'controlled' keys? If we supply tags with key "job" it will overwrite the job value in the metadata that the cli would otherwise provide.

tylerschultz:
After conversation w @cppforlife, we feel it would be better if the tags are not able to overwrite the bosh keys in the metadata. WDYT?

dpb587:
I had two use cases in mind: custom tags for IaaS/reporting/viewing; and replacing the hard-coded `bosh-init` value for `director` to match my environment name so it can show up in searches along with the rest of the environment (if I could rename the `director` key to `environment` I would).

I thought about only allowing `director` to be overridden since it was the only hard-coded value, but that would've been an extra few lines and also introduced the oddity that one internal field can be overridden but not others.

I am curious what the concern about values being overridden is though - do we have CPIs relying on the tag values, or are we planning on them doing that in the future? From the [possibly incorrect?] perspective that BOSH shouldn't care about those values, I see BOSH setting them as being more a default UX behavior. If a user wants something different and BOSH doesn't rely on the tags, it shouldn't care and I don't see a reason to prevent them. I also can't imagine users getting into a situation where they accidentally override the defaults and become surprised that `tags: { deployment: cf }` gives a tag value `cf` instead of `cf-a1b2c3d4`... the configuration is very intentional. And I don't think it's something that docs would need to recommend, per se; if necessary, docs say something like "the `director`, `deployment`, `job`, `index`, `id` tags are configured by default" and the user can do what they will.

So... I'm still a proponent of the PR as is, but happy to discuss more.

tylerschultz:
We're convinced.

dpb587:
I made this against `master` instead of `develop` :(

Sorry!

tjvman:
@dpb587 all good, `develop` is dead.

tjvman:
NO ITS NOT I LIED

tjvman:
Disregard any comments I made prior to this one.

dpb587:
Looks like it was eventually [merged](https://github.com/cloudfoundry/bosh-cli/commit/c19bcffad95940d155b4042e0d2720b156a683b5) back to master. I was confused when I saw master different than develop and then remembered I forgot to switch this branch. I guess a build is failing somewhere.

Carry on; don't trust my PRs and branch non-selections.

PR #160, opened 2017-03-06
Support using dynamic IP to recreate/upgrade director

zhanggbj:
Hi Dmitriy @cppforlife,

This is a commit to support using dynamic IP to recreate/upgrade director. Here are the background  and I addressed some of your comments after we discussed last time. Please help to take another review. Thank you for your time!

[#139519293] (https://www.pivotaltracker.com/story/show/139519293)

**[Background]**
Softlayer uses dynamic IP for deployment, but when to recreate/upgrade director(create-env in bosh-cli v2), we prefer its IP still remains the same. At the same time, Softylayer won’t delete the old VM and deploy a new one, it will do os-reload instead. So an property named “current_ip” is added to state.json file to indicate that using current IP as director IP address, not deleting the former VM. Then IP address is passed to create_vm Action and  bosh-softlayer-cpi will do the following os-reload accordingly.

/cc @maximilien @mattcui @jianqiu


cmd/deployment_deleter.go
```diff
@@ -100,6 +100,14 @@ func (c *deploymentDeleter) DeleteDeployment(stage biui.Stage) (err error) {
 		return bosherr.WrapError(err, "Loading deployment state")
 	}
 
+	if deploymentState.CurrentIP != "" {
```
cppforlife: i dont think we need to clear out current ip during deletion since we are deleting everything.

config/legacy_deployment_state_migrator.go
```diff
@@ -159,6 +160,7 @@ type instance struct {
 	UUID         string `yaml:"uuid"`
 	VMCID        string `yaml:"vm_cid"`
 	DiskCID      string `yaml:"disk_cid"`
+	DiskSize     int    `yaml:"disk_size"`
```
cppforlife: not sure how disk size relates to this PR?

deployment/vm/manager.go
```diff
@@ -154,3 +166,10 @@ func (m *manager) createAndRecordVM(agentID string, stemcell bistemcell.CloudSte
 
 	return cid, nil
 }
+
+func (m *manager) setVMNetworkInterfaces(vmNetworkInterfaces map[string]biproperty.Map, currentIP string, networkInterfaces map[string]biproperty.Map) {
+	for networkName, networkInterface := range networkInterfaces {
+		networkInterface["ip"] = currentIP
```
cppforlife: that's going to override ip field for all networks including VIP networks i believe.

cmd/deployment_deleter.go
zhanggbj: I added a check in [vm.go](https://github.com/zhanggbj/bosh-cli/blob/a3b247d4d946a071bc280ef0d1383a419ca8096d/deployment/vm/vm.go#L216) to ensure that If current_ip is found, it won’t delete vm , so bosh-softlayer-cpi could do os-reload after. When command `delete-env` is issued, this line ensure current_ip is "", so director is always deleted even it has been os-reloaded before(current_ip in not "").

deployment/vm/manager.go
zhanggbj: The IP address will be only used for dynamic network, as Softlayer doesn’t support VIP network currently. This will be handled in [bosh-softlayer-cpi create_vm](https://github.com/cloudfoundry/bosh-softlayer-cpi-release/blob/master/src/bosh-softlayer-cpi/softlayer/vm/softlayer_virtual_guest_creator.go#L44).

config/legacy_deployment_state_migrator.go
zhanggbj: The disk size is used for migrating from legacy deployment to v1 in Bluemix and I just double confirmed, this is not need anymore. So I’ll remove it in next patch.

zhanggbj:
Hi Dmitriy @cppforlife,

Thank you for your review. All comments addressed, please help to take another look. Thanks!

zhanggbj:
Hi Dmitriy @cppforlife,

In case you missed the notification:) I've addressed all of the comments, would you please help to take another review? Thank you for your time!

/cc @maximilien @mattcui

maximilien:
@zhanggbj thanks for the changes. Did @cppforlife schedule this in Tracker for the BOSH team?

deployment/vm/manager.go
cppforlife: @zhanggbj but this code runs for all IaaSes (break VIP networks).

zhanggbj:
Hi Dmitriy @cppforlife ,

These days we are refactoring bosh-softlayer-cpi code about os-reload and networks, so after a second thought, I think it's better to figure out another way based on the new code and resubmit.  I'll close this PR for now.

Thanks again for your review.

PR #121, opened 2017-02-09
Windows doesn't allow renaming files that are open

Fydon:
This change causes the unit tests to fail in Linux and Windows, but this function works on Windows with this change:

    Provider Get when URL starts with http(s):// when tarball is not present in cache when downloading succeds when saving to cache fails [It] returns an error 
    /tmp/build/1944c6a2/gopath/src/github.com/cloudfoundry/bosh-cli/installation/tarball/provider_test.go:161

      Expected
          <string>: Failed to download from 'http://fake-url': Saving downloaded bits to temporary file: write /tmp/temp-download-file428754441: bad file descriptor
      to contain substring
          <string>: fake-mkdir-error

      /tmp/build/1944c6a2/gopath/src/github.com/cloudfoundry/bosh-cli/installation/tarball/provider_test.go:160

      Full Stack Trace
        /tmp/build/1944c6a2/gopath/src/github.com/cloudfoundry/bosh-cli/vendor/github.com/onsi/gomega/internal/assertion/assertion.go:69 +0x1ed
      github.com/cloudfoundry/bosh-cli/vendor/github.com/onsi/gomega/internal/assertion.(*Assertion).To(0xc4203b3440, 0x8ec1c0, 0xc4203bb380, 0x0, 0x0, 0x0, 0xc4203bb380)
        /tmp/build/1944c6a2/gopath/src/github.com/cloudfoundry/bosh-cli/vendor/github.com/onsi/gomega/internal/assertion/assertion.go:35 +0xae
      github.com/cloudfoundry/bosh-cli/installation/tarball_test.glob..func2.2.2.3.3.5.2()
        /tmp/build/1944c6a2/gopath/src/github.com/cloudfoundry/bosh-cli/installation/tarball/provider_test.go:160 +0x230
      github.com/cloudfoundry/bosh-cli/vendor/github.com/onsi/ginkgo/internal/leafnodes.(*runner).runSync(0xc420203860, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
        /tmp/build/1944c6a2/gopath/src/github.com/cloudfoundry/bosh-cli/installation/tarball/suite_test.go:11 +0x64
      testing.tRunner(0xc42008a480, 0x7a0928)
        /usr/local/go/src/testing/testing.go:610 +0x81
      created by testing.(*T).Run
        /usr/local/go/src/testing/testing.go:646 +0x2ec

[...]

PR #171, opened 2017-03-12
Windows doesn't allow renaming files that are open (attempt 2)

Fydon:
This is a replacement for #121. All the tests now pass and the changes have been rebased to a recent version of the develop branch.


cppforlife:
@Fydon will pull it in.

installation/tarball/provider_test.go
```diff
@@ -77,18 +78,28 @@ var _ = Describe("Provider", func() {
...
 			Context("when tarball is not present in cache", func() {
 				var (
-					tempDownloadFilePath string
+					tempDownloadFilePath1 string
+					tempDownloadFilePath2 string
+					tempDownloadFilePath3 string
 				)
 
 				BeforeEach(func() {
-					tempDownloadFile, err := ioutil.TempFile("", "temp-download-file")
```
cppforlife: does this test fail on windows if not changed to have 3 temp files?

releasedir/fs_blobs_dir.go
```diff
@@ -244,6 +244,8 @@ func (d FSBlobsDir) TrackBlob(path string, src io.ReadCloser) (Blob, error) {
 
 	blobs[idx] = Blob{Path: path, Size: fileInfo.Size(), SHA1: sha1}
 
+	tempFile.Close()
```
cppforlife: any chance for a test for this one? nothing currently fails?

installation/tarball/provider_test.go
Fydon: This test fails everywhere now without 3 temp files, as the file is being closed on each [retry][1]. The real file system is creating a new temp file on each of the 3 retries, but the test was previously only using one temp file for all retries.

[1]: https://github.com/cloudfoundry/bosh-cli/pull/171/files/f78baee58902f252afde22804b19549c0618b805#diff-860904024a4888f514f430f2bffc5c41R120

releasedir/fs_blobs_dir.go
Fydon: Sure, but what should that test do? This method does [defer the close of the temp file][1], but now the defer close will fail if this line is reached. The defer close isn't checking for failure, so this close is causing a silent failure, but also preventing an error in Windows environments.

[1]: https://github.com/cloudfoundry/bosh-cli/pull/171/files/f78baee58902f252afde22804b19549c0618b805#diff-91f1299e44f3b411467c957c91b729bbL208

Fydon:
@cppforlife Should there be a defer close in `provider.go` to ensure that `downloadedFile` is always closed? In its current state could there be a resource leak?

Should I make defer close execute where I've added close? It seems that I could do this by splitting part of the function into another function or [using embedded functions][1]. The reason I ask this is that it might be confusing to have two close calls in the same function for the same resource, i.e. defer close and close.

[1]: https://play.golang.org/p/Xjy1g6W1Qf

Issue #177, opened 2017-03-20
deploy/int ignored ((xyz)) and didn't prompt for missing variable


drnic:
I placed some `((xyz))` variables in a deployment manifest and expected `bosh deploy manifest.yml` and `bosh int manifest.yml` to fail and complain the variables hadn't been provided.

Instead, they appeared in the manifest unmodified.

I'm aware of `--var-errs` flag; but perhaps it should be the default to fail on missing variables? This would help new users "fail fast" when not providing important variables; rather than booting up a system only to eventually debug and find `((xyz))` is a meaningless/erroneous input value.

pivotal-jamil-shamy:
@drnic having `--var-errs` as a default will present a problem when a user wants the director - not the cli - to interpolate the values using config server. 
In this case the CLI should act as a pipe passing the variables to the director, and it should not fail.

But to your point, I can see that it can get confusing to the user ( interpolating in the CLI and the director, perhaps both ).

drnic:
Couple of extra bits of info:* my directors in this case don't have a config-server so will not be interpolating any variables* these variables weren't defined in variable: section for interpolation by config-server
In either of these situations is there a reason for cli to not fail fast with missing variables?

cppforlife:
@drnic the reason why this is not currently happening in your director is because config server integration is not turned on (you can give it a try though through config-server or credhub ops files in bosh-deployment). since director is the end of the line for variable interpolation once we release config server integration officially, director will be always error before starting update procedure if it detects missing variables.

closing this since this is not a cli concern but director but feel free to reopen if plan is not clear.

drnic:
Sorry I'm confused a bit; perhaps there's an assumption I'm missing: is the cli assuming that all directors have a config- server? Is that currently true for anyone? (credhub-release still isn't publishing non-draft releases; config-server-release doesn't publish releases at all)
Second question: if a director did have a config-server - and I pass thru an unresolved variable that config- server isn't give a variables: to generate - then which part of the stack will eventually error?
Another way to ask: if there is ((xyz)) and no /variables/name=xyz/ then why wait for the config- server to fail: the cli knows in advance that it's not a variable that can be auto generated?
Sorry if I'm misunderstanding ((xyz)) with and without matching variables defn.

cppforlife:
> is the cli assuming that all directors have a config- server

it's for deploy and other director level commands because it will be default in the director at some point. dont want to add flags to cli that we will have to pull out in near future.

[...]

PR #178, opened 2017-03-20
Add deploy --var-errs

drnic:
`int --var-errs` is a helpful thing (though perhaps could be the default #177) but just discovered it's missing from `deploy`.

I think `deploy --var-errs` is important to ensure variables aren't missing. If its not the default, can it at least be added to `deploy` too please?

Update: PR attached


drnic:
OMG I copy and pasted like a champion.

cppforlife:
@drnic based on explanation i just added to https://github.com/cloudfoundry/bosh-cli/issues/177 im not too eager to pull this in.

PR #190, opened 2017-03-31
default to 'bosh' cli name; --with-bosh2 to name it bosh2

drnic:
This PR has not been tested. Sorry, not sure how.


cppforlife:
@drnic after chatting with a few people, we'll keep this until more people migrate to bosh v2 cli.

PR #191, opened 2017-04-05
Show cloud config diff before updating

voelzmo:
* Regenerates fakes before changing anything. Seems like this hasn't been done in a while. If you don't like this, we can skip the first commit.
* Adds an `ignoreNotFound` flag in `director/client_request`, which seems ugly, but we didn't find a good way to achieve backwards compatibility

### DIFF
--- a/cmd/update_cloud_config.go
+++ b/cmd/update_cloud_config.go
@@ -25,10 +25,31 @@ func (c UpdateCloudConfigCmd) Run(opts UpdateCloudConfigOpts) error {
 		return bosherr.WrapErrorf(err, "Evaluating cloud config")
 	}
 
+	cloudConfigDiff, err := c.director.DiffCloudConfig(bytes)
+	if err != nil {
+		return err
+	}
+
+	c.printManifestDiff(cloudConfigDiff)
+
 	err = c.ui.AskForConfirmation()
 	if err != nil {
 		return err
 	}
 
 	return c.director.UpdateCloudConfig(bytes)
+}
+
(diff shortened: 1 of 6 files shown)


director/client_request.go
```diff
@@ -106,11 +106,19 @@ func (r ClientRequest) RawGet(path string, out io.Writer, f func(*http.Request))
...
-	return r.readResponse(resp, out)
+	return r.readResponse(resp, out, false)
 }
 
 // RawPost follows redirects via GET unlike generic HTTP clients
 func (r ClientRequest) RawPost(path string, payload []byte, f func(*http.Request)) ([]byte, *http.Response, error) {
+	return r.rawPostInternal(path, payload, f, false)
+}
+
+func (r ClientRequest) RawPostIgnoreNotFound(path string, payload []byte, f func(*http.Request)) ([]byte, *http.Response, error) {
```
cppforlife: instead of adding a new method, RawPost already exposes http.Response. At the call site you can

```
if err != nil {
  if resp != nil && resp.StatusCode == 404 {
     logger.Info(logtag, "diffing of cloud config is not supported by the server")
    //skip error
  } else {
    return err
  }
}
```

beyhan:
@cppforlife I have updated the PR as suggested.

cppforlife:
@beyhan can you squash last 3 commits to make this a bit more readable. also can we remove regenerate fakes out of this pr?

beyhan:
@cppforlife We kicked out the generation and squashed the commits.

PR #207, opened 2017-04-27
Set disk metadata when attaching disk

mauromorales:
Bosh director introduced `set_disk_metadata` see: https://github.com/cloudfoundry/bosh/blob/master/src/bosh-director/lib/bosh/director/metadata_updater.rb#L32

With this PR we implemented the functionality inside the cli to use it when deploying.


deployment/vm/vm.go
```diff
@@ -151,6 +179,8 @@ func (vm *vm) AttachDisk(disk bidisk.Disk) error {
 		return bosherr.WrapError(err, "Attaching disk in the cloud")
 	}
 
+	vm.cloud.SetDiskMetadata(disk.CID(), vm.createDiskMetadata())
```
cppforlife: i take it we are ignoring error in this case to make this backwards compatible with older CPIs. let's log the error using logger.

Kiemes:
@cppforlife we made it backwards compatible by ignoring/logging the `NotImplemented` error. Any other error is returned.

beyhan:
We are still working on this to. It is not ready yet.

cppforlife:
@beyhan let us know when you want us to merge.

friegger:
@cppforlife the PR is ready for review

voelzmo:
Should we do https://www.pivotaltracker.com/story/show/144963353 in a separate commit before we release the CLI?

friegger:
@cppforlife @voelzmo We decided to do it all in one PR. Tagging with `instance_group` is now part of this PR to avoid dependencies between the two.

friegger:
Not ready yet.

cloud/errors.go
```diff
@@ -50,3 +55,33 @@ func (e cpiError) Message() string {
 func (e cpiError) OkToRetry() bool {
 	return e.cmdError.OkToRetry
 }
+
+func mapsToNotImplementedError(method string, cmdError CmdError) bool {
+	matched, _ := regexp.MatchString("^Method .+ not supported in photon CPI", cmdError.Message)
```
cppforlife: cant have `photon` naming in here.

cppforlife: may be just ignore any error for now? in 6 months we can be a bit more specific.

friegger: We would suggest to omit `photon`, e.g. `/^Method .+ not supported.*/`. We also need to keep this in sync with the director, where this has already been [implemented](https://github.com/cloudfoundry/bosh/blob/f40d35342ce8cd989ffa8eb73ef5045c2099ca36/src/bosh-director/lib/cloud/external_cpi.rb#L175).

We would really like to ignore only method missing style errors from cpis to avoid swallowing real errors. What is going to happen in 6 months?

manno:
It's ready now, we removed the Photon CPI specific error handling code.

PR #210, opened 2017-05-02
Homebrew formula defaults to `bosh` binary

wendorf:
Now that the CLI is GA, this should be the default.

Note: `build.without?` still works even though the property is `with`.

### DIFF
--- a/ci/tasks/update-homebrew-formula.sh
+++ b/ci/tasks/update-homebrew-formula.sh
@@ -17,7 +17,7 @@ class BoshCli < Formula
 
   depends_on :arch => :x86_64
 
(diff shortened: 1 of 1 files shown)


cppforlife:
@wendorf chatting with a few people, i think we should keep this for a little bit as is until more people flip over to the new cli.

wendorf:
Sounds good to me. Can we keep the story/PR open and maybe prioritized for
a couple iterations away? I'd rather push it down a couple times than
forget about it.

cppforlife:
closing in favor of https://github.com/cloudfoundry/bosh-cli/pull/190. that PR has code that is necessary for the actual change.

PR #212, opened 2017-05-03
Fix typo for failing task

bkrannich:


cppforlife:
@bkrannich there are a few references to that text in the repo (searched for "to succeed but state").

bkrannich:
@cppforlife: Yes, sorry about that. Fixed the corresponding test with fa06f0e. The other search hits have a different context/don't contain the typo.

Issue #31, opened 2016-10-24
Weird line-break in 'cck' output


voelzmo:
Here is what I found doing `bosh cck` recently. It's breaking right in the middle of the word 'cloud', which seems totally weird:

```
$ bosh-go cck -d dummy
Using environment '172.18.104.90' as user 'admin'

Using deployment 'dummy'

Task 6968

09:00:08 | Scanning 1 VMs: Checking VM states (00:00:12)
09:00:20 | Scanning 1 VMs: 0 OK, 0 unresponsive, 1 missing, 0 unbound (00:00:00)
09:00:20 | Scanning 1 persistent disks: Looking for inactive disks (00:00:03)
09:00:23 | Scanning 1 persistent disks: 1 OK, 0 missing, 0 inactive, 0 mount-info mismatch (00:00:00)

Started  Mon Oct 24 09:00:08 UTC 2016
Finished Mon Oct 24 09:00:23 UTC 2016
Duration 00:00:15

Task 6968 done

#    Type        Description
105  missing_vm  VM for 'dummy_with_package_z1/dcbbf18c-b2f8-482d-a58e-f1297841300f (0)' with cloud ID '07892183-53e9-4e2e-9620-a546b59cbeab' missing.

1 problems

1: Skip for now
2: Recreate VM without waiting for processes to start
3: Recreate VM and wait for processes to start
4: Delete VM reference
VM for 'dummy_with_package_z1/dcbbf18c-b2f8-482d-a58e-f1297841300f (0)' with clo
ud ID '07892183-53e9-4e2e-9620-a546b59cbeab' missing. (1):
```

Here is my CLI version

```
$ bosh --version
version 0.0.87-88598e3-2016-10-20T02:36:56Z

Succeeded
```

cppforlife:
@voelzmo sure you didnt press ENTER at that time?

@vito any reason why go-interact would wrap? term width?

vito:
I've seen this too, not wrapping at the prompt but wrapping on user input weirdly. Maybe something changed in a dependency of go-interact?

voelzmo:
@cppforlife definitely didn't press ENTER at that time, this happens every time I call it. ;)

PR #220, opened 2017-05-12
Bump github.com/vito/go-interact/interact

dpb587:
* fixes extra newlines in longer prompts
 * fixes cloudfoundry/bosh-cli#31

Full vendor diff at [0eb3903...c573392](https://github.com/vito/go-interact/compare/0eb3903...c573392); only skimmed it

Manually verified; lightly tested other commands; units/integration green; acceptance too confusing


voelzmo:
nice catch, thanks for the fix!

PR #222, opened 2017-05-12
Make SSH username compatible with Windows

charlievieth:
Windows has a max username length of 20 chars, the bosh-cli was
generating a username 21 chars long, this commit reduces the username
length by 1 to 20.

@natalieparellano 

[#145070025]

Signed-off-by: Charlie Vieth <cviethjr@pivotal.io>

### DIFF
--- a/director/ssh_opts_test.go
+++ b/director/ssh_opts_test.go
@@ -3,6 +3,8 @@ package director_test
 import (
 	"crypto/rsa"
 	"errors"
+	"fmt"
+	"strings"
 
 	fakeuuid "github.com/cloudfoundry/bosh-utils/uuid/fakes"
 	. "github.com/onsi/ginkgo"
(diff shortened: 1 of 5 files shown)


director/ssh_opts_test.go
```diff
@@ -13,21 +15,43 @@ import (
...
 	BeforeEach(func() {
 		uuidGen = &fakeuuid.FakeGenerator{
-			GeneratedUUID: "2a4e8104-dc50-4ad7-939a-2efd53b029ae",
+			GeneratedUUID: UUID,
+		}
+	})
+
+	It("generates a username that is compatible with Windows", func() {
+		const MaxLength = 20
+		const InvalidChars = "\"/\\[]:|<>+=;?*"
```
dpb587-pivotal: Nice additional test coverage. If there's an official list of invalid chars somewhere, it'd be helpful to include a comment/link for more documentation.

dpb587-pivotal:
Long-term, I think it'd be nice if director or agent should be generating ephemeral user IDs instead of users having that sort of control. But that's a different story.

director/ssh_opts_test.go
charlievieth: Good call, [heres the doc](https://technet.microsoft.com/en-us/library/cc771865(v=ws.11).aspx#Parameters), we'll make a PR to add it.

PR #224, opened 2017-05-15
Add `instance_group` to VM Metadata

friegger:
Also adding `instance_group` for disk metadata will be done in #207 

[#144963353](https://www.pivotaltracker.com/story/show/144963353)


friegger:
Handled in #207

PR #230, opened 2017-05-22
Support diff for update-runtime-config command

friegger:
https://www.pivotaltracker.com/story/show/141968405


cmd/print_diff.go
```diff
@@ -0,0 +1,19 @@
+package cmd
```
dpb587-pivotal: Seems a bit out of place to have this as a standalone in `cmd` package. Probably not necessary to move it at this point, but I wonder if it might make more sense in a subpackage of `ui` someday. dk probably has opinions on that.

cmd/opts.go
```diff
@@ -328,7 +328,8 @@ type UpdateRuntimeConfigOpts struct {
 	VarFlags
 	OpsFlags
 
-	Name string `long:"name" description:"Runtime-Config name (default: '')" default:""`
+	NoRedact bool   `long:"no-redact" description:"Show non-redacted manifest diff"`
```
dpb587-pivotal: Should add a new test in [`opts_test.go`](https://github.com/cloudfoundry/bosh-cli/blob/35a76411753f85311e82351f1906510f1c494e3a/cmd/opts_test.go#L1089).

cmd/print_diff.go
cppforlife: i think it's ok to have it in cmd package, but let's make it an object, like Diff which would have Print method.

loewenstein:
Closed in favor of https://github.com/cloudfoundry/bosh-cli/pull/234

PR #231, opened 2017-05-24
Show cpi config diff when updating cpi config

mauromorales:
We've added support to show cpi diffs on bosh director: https://github.com/cloudfoundry/bosh/commit/fe7c2e2dd7dfcf63ebdadb9dad30b1aac913947b

With this PR we add cpi config diffing support to the cli as well.

[#141968471](https://www.pivotaltracker.com/story/show/141968471)

### DIFF
--- a/cmd/opts.go
+++ b/cmd/opts.go
@@ -311,6 +311,9 @@ type UpdateCPIConfigOpts struct {
 	Args UpdateCPIConfigArgs `positional-args:"true" required:"true"`
 	VarFlags
 	OpsFlags
+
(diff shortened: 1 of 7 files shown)


luan:
Hi @mauromorales @Kiemes,

It looks like there was something missing from this PR, we got
```
# github.com/cloudfoundry/bosh-cli/cmd
cmd/update_cpi_config.go:33: undefined: printManifestDiff
```
so we reverted this merge for now so that you can fix and push again ok?

Thanks,
@luan and @baguasquirrel

PR #157, opened 2017-03-02
WIP: follow symlinks on create-release

ljfranklin:
- capi-release and others make use of symlinks to share templates
  between multiple jobs, e.g. https://github.com/cloudfoundry/cloud_controller_ng/blob/3db08e602ef4f7156c681a8799bc53a2c6e65848/bosh/shared_job_templates/setup_local_blobstore.sh.erb
- While some of our shared templates could be moved into packages or
  links, there are some files that don't fit this pattern like scripts that
  require ERB interpolation
- TODO: add tests

PR #232, opened 2017-05-24
Follow symlinks in jobs

jenspinney:
This PR implements symlink-following (i.e., copying the contents of the target of the symlink) in the jobs folder when creating a release.

- capi-release and others make use of symlinks to share templates between multiple jobs, e.g., https://github.com/cloudfoundry/cloud_controller_ng/blob/3db08e602ef4f7156c681a8799bc53a2c6e65848/bosh/shared_job_templates/setup_local_blobstore.sh.erb
- While some of our shared templates could be moved into packages or links, there are some files that don't fit this pattern like scripts that require ERB interpolation.
- Some bosh users want to copy the symlink itself (not the target) from a package onto the BOSH VM, so we restrict this symlink-following behavior to only happen when processing the jobs directory.

Verification: We were able to create and upload capi-release without any errors using this version of the CLI. We additionally added a symlink inside one of the capi-release packages which pointed to /var/vcap/sys/logs, which doesn't exist on our development system. When we deployed the release, all the jobs started successfully and we verified that our logs symlink correctly pointed to /var/vcap/sys/logs on the bosh VM.

This completes the work of the WIP PR #157 and implements story [#145780425](https://www.pivotaltracker.com/story/show/145780425) in the capi tracker.

-- @jenspinney and @mikexuu

### DIFF
--- a/release/resource/archive.go
+++ b/release/resource/archive.go
@@ -26,11 +26,8 @@ type ArchiveImpl struct {
 }
 
 func NewArchiveImpl(
-	files []File,
-	prepFiles []File,
-	additionalChunks []string,
+	args ArchiveFactoryArgs,
 	releaseDirPath string,
-	followSymlinks bool,
 	fingerprinter Fingerprinter,
 	compressor boshcmd.Compressor,
 	digestCalculator bicrypto.DigestCalculator,
@@ -38,11 +35,11 @@ func NewArchiveImpl(
 	fs boshsys.FileSystem,
 ) ArchiveImpl {
 	return ArchiveImpl{
-		files:            files,
-		prepFiles:        prepFiles,
-		additionalChunks: additionalChunks,
+		files:            args.Files,
+		prepFiles:        args.PrepFiles,
+		additionalChunks: args.Chunks,
+		followSymlinks:   args.FollowSymlinks,
 		releaseDirPath:   releaseDirPath,
-		followSymlinks:   followSymlinks,
 
 		fingerprinter:    fingerprinter,
 		compressor:       compressor,
@@ -211,5 +208,5 @@ func (a ArchiveImpl) buildStagingArchive(stagingDir string) Archive {
 	}
 
 	// Initialize with bare minimum deps so that fingerprinting can be performed
-	return NewArchiveImpl(stagingFiles, nil, a.additionalChunks, "", false, a.fingerprinter, nil, nil, nil, nil)
+	return NewArchiveImpl(ArchiveFactoryArgs{Files: stagingFiles, Chunks: a.additionalChunks}, "", a.fingerprinter, nil, nil, nil, nil)
 }
--- a/release/resource/archive.go
+++ b/release/resource/archive.go
@@ -162,23 +162,23 @@ func (a ArchiveImpl) copyFile(sourceFile File, stagingDir string) error {
 		return err
 	}
 
+	sourceFilePath := sourceFile.Path
 	if sourceFileStat.Mode()&os.ModeSymlink != 0 {
-		symlinkTarget, err := a.fs.Readlink(sourceFile.Path)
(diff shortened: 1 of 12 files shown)


cppforlife:
@jenspinney given that bools are very descriptive it would be nice to add an options struct in place of where bool is accepted. struct could have FollowSymlinks field that will make the initialization a bit more understandable.

jenspinney:
Yeah, that's a good idea, @cppforlife. Our latest commit has a struct for the args to archiveFactory instead of a list of parameters.

release/resource/archive.go
```diff
@@ -29,6 +30,7 @@ func NewArchiveImpl(
 	prepFiles []File,
 	additionalChunks []string,
 	releaseDirPath string,
+	followSymlinks bool,
```
cppforlife: i think it's worth changing places like this to struct as well.

jenspinney:
Ok thanks for the feedback @cppforlife! We added a commit which passes the struct along in the NewArchiveImpl() method as well.

- @jenspinney && @anniesing

rosenhouse:
@cppforlife We just hit the issue with `bosh-cli` where we were unable to build `capi-release` because it does not follow symlinks.  I don't see any tracker integration here, is this PR scheduled to get merged at some point?

cppforlife:
@rosenhouse yes, we'll be merging this in shortly.

baguasquirrel:
@jenspinney, @rosenhouse, @cppforlife 

So, we tried to merge this PR and ran into issues when we tried to run the unit tests in our CI, from the bosh-cli repo:

`fly -t <concourse_alias> execute -c ci/tasks/test-unit.yml -i bosh-cli=.`

The new tests that were written don't work under Linux. Specifically, it fails on the file permissions check in archive_test.go:323, "when following symlinks." The test expects directory permissions 0755 but under linux it gets 0777. We looked into the code in archive.go, dir_reader.go, and fingerprint.go, and tried to see if we should/could use the permissions of symlinked target, but we found numerous comments explaining why it was hardcoded to 0755. The file permissions definitely need to be 0755, so your original test was correct. Something about the implementation isn't working on Linux, and we're punting it back to you.

jenspinney:
@baguasquirrel @cppforlife  As far we as could tell, the 0755 restriction mentioned in the comments was for files only, not directories (as the existing code just directly copies the permissions from a source directory when it creates a new directory, without changing it to 0755, for example).

The old test actually had a bug where it was relying on the umask being set to a particular value in order for the test to pass, and this is why there was a difference between linux and MacOS for the new test we added. We hardcoded the umask to be 0 in order to ensure test consistency across systems and environments. Setting the umask to 0 revealed that the "dir" directory being copied (which has nothing to do with symlinks) was only relying on the test environment's umask to set the permissions to 0755 -- nothing in the code actually does that.

So from what we could see, in the existing implementation, only non-directory files have this 0755 or 0644 rule applied to them. In the existing code, directory permissions get copied exactly as-is. It only looked like directory permissions were being changed to 0755 because the umask wasn't 0. We've fixed the follow-symlink implementation to continue this pattern of copying the directory permissions as-is. In the case of followed symlinks, that means we copied the permssions of the target directory.

If this is a problem, please let us know, but it seems like what we're doing now is at least consistent with existing behavior for directories.

Thanks,
@jenspinney && @ericpromislow

tjvman:
Merged into develop as commit cb775570a0dcf3c6dd7455f318c04cc28a69eccd

tjvman:
@jenspinney In the future, please submit pull requests for the cli to develop, rather than master. Manually pulled this one into develop.

jenspinney:
Thanks @tjvman! I think you can configure the default branch for PRs. Right now it's `master`, so I assumed that was the right one. You may want to change this so others don't make the same mistake.

PR #233, opened 2017-05-26
Upload binaries to GitHub release

### LINKED ISSUE #187: Please upload tarballs of actual bosh CLI to your github release
Would make it easier than having to go to S3 and finding the tarball.

dpb587:
If desired. Seems like it should be done as a courtesy for users and automation. If they want to be secure they should be checking from bosh.io with the checksum.

May want to consider enabling `drafts` to avoid publishing a release for every build.

Requires a new variable of `concourse_github_access_token` with a GitHub token which can create releases.

Fixes #187

### DIFF
--- a/ci/pipeline.yml
+++ b/ci/pipeline.yml
@@ -174,6 +174,16 @@ jobs:
             tag: version-semver/number
             tag_as_latest: true
 
+      - put: github-release
+        params:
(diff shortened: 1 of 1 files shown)


jaresty:
merge story [#152974208](https://www.pivotaltracker.com/story/show/152974208)

PR #234, opened 2017-05-29
Runtime config diff

mauromorales:
supersedes https://github.com/cloudfoundry/bosh-cli/pull/230

https://www.pivotaltracker.com/story/show/141968405


cmd/print_diff.go
```diff
@@ -0,0 +1,29 @@
+package cmd
+
+import (
+	boshui "github.com/cloudfoundry/bosh-cli/ui"
+)
+
+type Diff struct {
+	Lines [][]interface{}
```
cppforlife: `lines` could be private.

mauromorales:
@cppforlife changed the lines field to be unexported in the Diff struct. Let me know if there is anything else

PR #238, opened 2017-06-08
Introduce GCSBlobstore with provider hook

Everlag:
GCSBlobstore introduces initial support for native
GCS blobstore support. We use bosh-gcscli/client as
our client package.
https://github.com/cloudfoundry/bosh-gcscli

This mirrors the existing patterns of S3Blobstore.
Specifically, we leave Delete unimplemented to keep parity.

Significant portions of S3Blobstore are duplicated due to an
SDK-level compatibility issue. Specifically, s3cli/client.Get
expects an io.WriterAt as the SDK accepts that. The GCS SDK
accepts only io.Writer. Hence, we cannot share the code by
having a common interface for bosh-gcscli/client and s3cli/client.
The other option is using interface{} and a type-switch in
each Blobstore method wherever the client is used. The result
of this was undesirable.

### DIFF
--- a/releasedir/gcs_blobstore.go
+++ b/releasedir/gcs_blobstore.go
@@ -1,0 +1,113 @@
+package releasedir
+
+import (
+	gobytes "bytes"
+	"context"
+	"encoding/json"
+	"os"
+
+	bosherr "github.com/cloudfoundry/bosh-utils/errors"
+	boshsys "github.com/cloudfoundry/bosh-utils/system"
+	boshuuid "github.com/cloudfoundry/bosh-utils/uuid"
+
+	gcsclient "github.com/cloudfoundry/bosh-gcscli/client"
+	gcsconfig "github.com/cloudfoundry/bosh-gcscli/config"
+)
+
+type GCSBlobstore struct {
+	fs      boshsys.FileSystem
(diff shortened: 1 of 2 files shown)


releasedir/gcsblobstore.go
```diff
@@ -0,0 +1,114 @@
...
+import (
+	gobytes "bytes"
+	"context"
+	"os"
+
+	bosherr "github.com/cloudfoundry/bosh-utils/errors"
+	boshsys "github.com/cloudfoundry/bosh-utils/system"
+	boshuuid "github.com/cloudfoundry/bosh-utils/uuid"
+
+	"encoding/json"
```
cppforlife: move up with other std deps?

Everlag: Done.

cppforlife:
left two issues before we merge this one in:

- https://github.com/cloudfoundry/bosh-gcscli/issues/11
- https://github.com/cloudfoundry/bosh-gcscli/issues/12

Everlag:
Renamed to gcs_blobstore and resolved both of those issues.

dpb587-pivotal:
Thanks!

PR #241, opened 2017-06-14
Show vm_created_at for instances and vms with vitals option

barthy1:
[#146195037](https://www.pivotaltracker.com/story/show/146195037)


barthy1:
Related to https://github.com/cloudfoundry/bosh/pull/1707

cmd/instance_values.go
```diff
@@ -30,6 +31,10 @@ type ValueUptime struct {
 	Secs *uint64
 }
 
+type ValueVMCreatedAt struct {
```
cppforlife: instead of introducing VMCreatedAt lets change ValueTime to have String() return "" if time equals epoch (should be zero value of time.Time?). also let's have director return iso8601 string instead of an int64 to be more consistent with the rest of APIs (similar to https://github.com/cloudfoundry/bosh-cli/blob/master/director/orphan_disks.go#L65; use TimeParser object?).

PR #249, opened 2017-06-29
log-out command requires the environment to be set

apoydence:
This prevents a user from doing `bosh log-out` without providing an environment.


apoydence:
@cppforlife Any chance this could be merged?

dpb587-pivotal:
Useful - thanks!

PR #255, opened 2017-07-06
Fix gcs vendoring

loewenstein:
Didn't test this version. But at least bin/build did work again


loewenstein:
Just seen I had my trivial workaround for having some cli from develop still in the repo.

PR #256, opened 2017-07-07
Support optional GOPATH in update-dep script

alex-slynko:


cppforlife:
we have started using golang/dep in develop yesterday since we had trouble packaging bosh-gcscli. works much better.

PR #258, opened 2017-07-09
Properly format JSON output for ValueInt and ValueInteger

### LINKED ISSUE #257: JSON output of `inspect-release` is difficult to parse
When parsing the `links_consumed` and `links_provided` from:
```
bosh -e lite inspect-release myrelease/0+dev.1 --json
```

The resulting output has:
```
"links_consumed": "- name: myjob\n  type: myjob\n  optional: true",
"links_provided": "- name: myotherjob\n  type: myotherjob"
```

I would have expected this to be better formatted. As an example:
```
"links_consumed": [{
  "name": "myjob",
  "type": "myjob",
  "optional": "true"
}], 
"links_provided": [{
  "name": "myotherjob",
  "type": "myotherjob"
}]
```

I'm running the bosh-cli version: 
```
bosh --version
version 2.0.26-105157c-2017-06-22T18:54:04Z
```

apoydence:
resolves #257

### DIFF
--- a/ui/json_ui.go
+++ b/ui/json_ui.go
@@ -28,7 +28,7 @@ type uiResp struct {
 type tableResp struct {
 	Content string
 	Header  map[string]string
-	Rows    []map[string]string
+	Rows    []map[string]interface{}
 	Notes   []string
 }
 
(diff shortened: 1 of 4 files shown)


ui/json_ui.go
```diff
@@ -134,18 +134,25 @@ func (ui *jsonUI) Flush() {
...
+		data := map[string]interface{}{}
 
 		for i, col := range row {
-			if header[i].Hidden {
+			if header[i].Hidden || col.String() == "" {
 				continue
 			}
 
-			data[header[i].Key] = col.String()
+			switch v := col.(type) {
```
cppforlife: instead of doing it based on concrete type, let's add `Raw() interface{}` to each value type, then here we can just cast to this interface. if not castable to interface, fallback to String().

apoydence: I've updated the commit.

apoydence:
@cppforlife Do the changes look the way you were wanting?

cppforlife:
@apoydence we are in the process of updating integration tests to fit the new output. it's quite a bit change so taking a few days.

jaresty:
merge story [#152974024](https://www.pivotaltracker.com/story/show/152974024)

PR #259, opened 2017-07-12
make bosh create-release error out with hint to `reset-release` when old-style blobs dir is present

mauromorales:
If you run `create-release` with bosh-cli v2 on a release that has previously generated blobs from bosh-cli v1 then cli will exit and hint you to run `reset-release`.

There is a conversation in [slack](https://cloudfoundry.slack.com/messages/C094QRLEQ/convo/C094QRLEQ-1499156841.349691/) if you want to follow up.

[#145341545](https://www.pivotaltracker.com/story/show/145341545)

### DIFF
--- a/releasedir/fs_blobs_dir.go
+++ b/releasedir/fs_blobs_dir.go
@@ -312,8 +312,7 @@ func (d FSBlobsDir) containsSymlinks() bool {
 
 	for _, file := range files {
 		fileInfo, _ := d.fs.Lstat(file)
-
-		if fileInfo.Mode()&os.ModeSymlink != 0 {
+		if fileInfo != nil && fileInfo.Mode()&os.ModeSymlink != 0 {
 			return true
 		}
 	}
--- a/releasedir/fs_blobs_dir.go
+++ b/releasedir/fs_blobs_dir.go
@@ -118,6 +118,10 @@ func (d FSBlobsDir) SyncBlobs(numOfParallelWorkers int) error {
 		return err
 	}
 
+	if d.containsSymlinks() {
+		return bosherr.Error("Bailing because symlinks found in blobs directory. If switching from CLI v1, please use the `reset-release` command.")
+	}
+
 	if err := d.removeUnknownBlobs(blobs); err != nil {
 		return bosherr.WrapErrorf(err, "Syncing blobs")
 	}
@@ -298,6 +302,23 @@ func (d FSBlobsDir) UploadBlobs() error {
 	}
 
 	return nil
+}
+
+func (d FSBlobsDir) containsSymlinks() bool {
+	files, err := d.fs.RecursiveGlob(filepath.Join(d.dirPath, "**/*"))
+	if err != nil {
+		return false
+	}
+
+	for _, file := range files {
+		fileInfo, _ := d.fs.Lstat(file)
+
+		if fileInfo.Mode()&os.ModeSymlink != 0 {
+			return true
+		}
+	}
+
+	return false
 }
 
 func (d FSBlobsDir) checkBlobExistence(dstPath string, digest boshcrypto.MultipleDigest) bool {
--- a/releasedir/fs_blobs_dir.go
+++ b/releasedir/fs_blobs_dir.go
@@ -118,7 +118,12 @@ func (d FSBlobsDir) SyncBlobs(numOfParallelWorkers int) error {
 		return err
 	}
 
-	if d.containsSymlinks() {
+	symlinksFound, err := d.containsSymlinks()
+	if err != nil {
+		return bosherr.WrapErrorf(err, "Syncing blobs")
+	}
+
+	if symlinksFound {
 		return bosherr.Error("Bailing because symlinks found in blobs directory. If switching from CLI v1, please use the `reset-release` command.")
 	}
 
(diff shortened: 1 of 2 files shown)


releasedir/fs_release_dir.go
```diff
@@ -206,6 +206,10 @@ func (d FSReleaseDir) BuildRelease(name string, version semver.Version, force bo
 		return nil, err
 	}
 
+	if d.blobsDir.ContainsSymlinks() {
+		return nil, bosherr.Error("Execute `bosh-release` command to clean up any blobs artifacs left from using the bosh-cli v1.")
```
dpb587-pivotal: Can this message be fixed and cleaned up and perhaps suggest why it looks wrong? Perhaps

    Bailing because symlinks found in blobs directory. If switching from CLI v1, please use the `reset-release` command.

mauromorales:
@dpb587 thanks, we updated it with the suggested message.

release/dir_reader.go
```diff
@@ -8,6 +8,7 @@ import (
 	boshlog "github.com/cloudfoundry/bosh-utils/logger"
 	boshsys "github.com/cloudfoundry/bosh-utils/system"
 
+	"fmt"
```
cppforlife: std deps go to the top.

release/release.go
```diff
@@ -158,6 +159,7 @@ func (r *release) Build(devIndicies, finalIndicies ArchiveIndicies) error {
 	}
 
 	for _, pkg := range r.Packages() {
+		fmt.Printf("release.Build %v %v\n", pkg, devIndicies.Packages)
```
cppforlife: dead code.

releasedir/fs_blobs_dir.go
```diff
@@ -300,6 +302,24 @@ func (d FSBlobsDir) UploadBlobs() error {
 	return nil
 }
 
+func (d FSBlobsDir) ContainsSymlinks() bool {
+	files, err := d.fs.RecursiveGlob(filepath.Join(d.dirPath, "**/*"))
+
```
cppforlife: empty space line

releasedir/interfaces.go
```diff
@@ -73,6 +73,8 @@ type BlobsDir interface {
 
 	TrackBlob(string, io.ReadCloser) (Blob, error)
 	UntrackBlob(string) error
+
+	ContainsSymlinks() bool
```
cppforlife: may be instead of exposing it as a public method, it makes more sense to call it from inside SyncBlobs()? /cc @dpb587

mauromorales:
@dpb587-pivotal we had some trouble pushing yesterday, we've now cleaned up the commits, squashed and force pushed. As suggested by @cppforlife we also don't export `ContainsSymlinks()` and instead call it from within `SyncBlobs()`

releasedir/interfaces.go
dpb587: Agreed it shouldn't be part of the public interface.

releasedir/fs_blobs_dir.go
```diff
@@ -300,6 +304,23 @@ func (d FSBlobsDir) UploadBlobs() error {
...
 }
 
+func (d FSBlobsDir) containsSymlinks() bool {
+	files, err := d.fs.RecursiveGlob(filepath.Join(d.dirPath, "**/*"))
+	if err != nil {
+		return false
+	}
+
+	for _, file := range files {
+		fileInfo, _ := d.fs.Lstat(file)
```
cppforlife: https://golang.org/pkg/os/#FileInfo is an interface. before using it, have to check for err.

releasedir/fs_blobs_dir_test.go
```diff
@@ -416,6 +416,47 @@ bad-sha-blob.tgz:
 				Expect(err).To(MatchError("Syncing blobs: Removing unknown blob: failed to remove " + filepath.Join("/", "dir", "blobs", "extra-blob.tgz")))
 			})
 		})
+
+		Context("when a symlink exists", func() {
+			It("returns an error", func() {
+
```
cppforlife: remove empty line.

releasedir/fs_blobs_dir_test.go
```diff
@@ -416,6 +416,47 @@ bad-sha-blob.tgz:
...
+					},
+				})
+
+				err := act(1)
+				Expect(err).To(MatchError("Bailing because symlinks found in blobs directory. If switching from CLI v1, please use the `reset-release` command."))
+			})
+		})
+
+		Context("when no symlink exists", func() {
+			It("returns false", func() {
```
cppforlife: "returns false" -> "succeeds"

releasedir/fs_blobs_dir.go
dpb587-pivotal: Seems like we're still ignoring the error here; is that intentional?

Given we know there's a file there, it seems like we should be able to expect a successful Lstat call. Anything else suggests an fs issue that we probably should be concerned with.

mauromorales: @dpb587-pivotal originally I was thinking that unless I can verify the FileInfo of the symlink I cannot really react on the issue but you are right I modified the function to also return an error now and pass along the PathError if it exists.

PR #263, opened 2017-07-14
Add JQ

zaksoup:


cppforlife:
@zaksoup fixed in https://github.com/cloudfoundry/bosh-cli/commit/1f8ec439c59fa5451bd6ad5f8923b9ea9a84755c

PR #269, opened 2017-07-20
Introduce Source Repo Url for releases

petergtz:
For more context, please see: https://github.com/cloudfoundry/bosh/issues/1525


cppforlife:
@petergtz may be making this a bit more generic would be better? something like `metadata` hash that would be sibling to name and version. inside of it could have `source_url`, i could see also may be adding an author and other misc data?

```
name: foo
version: 123
metadata:
  source_url: https://github.com/cloudfoundry/foo
```

@dpb587-pivotal you've thought i think a little bit about this before. wdyt?

petergtz:
@cppforlife Makes sense. In the code, would you treat `metadata` as a typed struct, something along the lines of:
```go
type ReleaseMetadata struct {
    SourceUrl string
    Author string
}
```

Or would you treat it as a `map[string]string`?

I can see both possibilities, but I'd like us to consider it in the broader context of https://github.com/cloudfoundry/bosh/issues/1525, i.e. in the end I'd like us to be able to do:

```diff
$ bosh -d my-deployment diff-deployment-revisions --include-git-diff 3 4

# manifest-diff would go here...

diff --git a/jobs/bits-service/spec b/jobs/bits-service/spec
index 4aab31f..0b4cba3 100644
--- a/jobs/bits-service/spec
+++ b/jobs/bits-service/spec
@@ -26,7 +26,7 @@ packages:

 properties:
   request_timeout_in_seconds:
-    description: "Timeout for requests in seconds."
+    description: "The changed description"
     default: 900

   bits-service.public_endpoint:
```

And to be able to do that, the bosh CLI would have to extract `source_url` and `commit_hash` and do a `git diff` (obviously it will refuse to do that when the release wasn't created with these values in the first place). So I'm wondering if we should make these values a first class concept, or if we can still live with them being an untyped hash.

What do you think?

dpb587-pivotal:
I like the metadata idea (and having repository URI accessible somewhere).

I don't think the repository URL should be committed based on the local repo creating the final release though. Releases might often be committed in local forks before being pushed externally, and that would result in releases having a possibly incorrect/`file://` repo.

In prior experiments, I've used `config/final.yml` as a place to commit additional metadata. For example, if there was a `metadata` field in there, `create-release` could potentially copy that into the finalized manifest. This would allow for arbitrary metadata to be committed and versioned like other release configuration items (I've also used it for URLs, license summary, meta-resources for other tooling, and signing information).

Naming it `metadata` in `config/final.yml` might be a bit too generic, but I think `metadata` would fit inside a final release YAML. Perhaps there's a different name which could be shared in both places. I'd imagine it more of a generic `map[string]interface{}`.

From the CLI, perhaps metadata information could be exposed as part of the `inspect-release` command. Maybe using `--metadata`. It'd require a new field in the database for recording the new metadata.

If core director starts depending on the repository value, it might be better to make it a more official property (rather than in metadata; I still think it should be derived from a `config/final.yml` rather than from local `create-release` env).

Regardless, I think a generic `metadata` feature would be useful and it makes it easier to prototype these sorts of experimental ideas to see if they should become more official "data".

Not sure what @cppforlife thinks.

petergtz:
@dpb587-pivotal I understand your concern with deriving the source URL automatically from the git remote. But I think we are looking at it from two different perspectives. From my perspective, the main use case of this is that Concourse (or any other automation pipeline) clones the git repo, creates a release and uploads it somewhere. Just like you don't have to care about the commit hash you wouldn't care about the source URL and it comes for free. Of course (and my PR does not include that), the `create-release` command could have a new option to override that source URL in case the git remote for whatever reason is non-sensical.

The good thing about this approach is that you only need a new version of the BOSH CLI, and to actually store it a new version of the bosh director. Both happens sooner or later in any environment, and you'd consume this new feature automatically, whereas your suggestion requires you to do this to every existing BOSH release individually. So I think adoption rate would suffer a lot. In fact, I doubt it would ever be used in certain environments (despite the fact that it would benefit those environments).

Also, I think when having a local fork that points to `file:///` at the time of release creation, it really just manifests your release creation process in the manifest which I believe is a good thing. Maybe it's a motivator to automate that step. If not, you can still override it using an option during `create-release`.

I was also wondering: if this value lives in `final.ym`, isn't there a real danger that this value gets stale and eventually the commit hash might not even be existent in the repo pointed to by the source URL? Or the URL simply doesn't exist anymore. I'm having moves in mind, like from `github.com/cloudfoundry-incubator` to `github.com/cloudfoundry`.

In your suggestion, would `dev_releases` also grab the source URL from `final.yml`? (I think dev_release should also contain a source URL in their manifest, because in dev environment it's just as useful to see which diff broke the acceptance tests.)

As you can see, my main concern is about actually having a working source URL in **every** BOSH release, so we could introduce release diffs as I described in my earlier comment. I want us to be able to push several new releases into Production a day, while still maintaining full control of what changed when. 

So in summary, I think `git remote origin` would provide a sane default and could be easily overridden when there's need. What do you think?

/cc @cppforlife

cppforlife:
@petergtz hey, ive chatted with @dpb587 and @tjvman more. let's expand this pr to do following things:

- allow `repository`, `license`, `description` as top level keys in final.yml and release.MF (similar to https://github.com/lynnux/xpsupport/blob/master/Cargo.toml#L8)
- fill in repository automatically if not present

separately director should add 3 columns to release_versions (next to commit_hash).

wdyt?

petergtz:
@cppforlife Sounds good! Will start working on it.

petergtz:
@cppforlife I realized there is already logic that copies any `LICENSE` or `NOTICE` file that exists in the release dir into release itself. Does it still make sense to have an extra field in the manifest for license? How will we deal with conflicting data? Is that simply the release-authors responsibility?

cppforlife:
>  Is that simply the release-authors responsibility?

yeah, let's go with that for now.

petergtz:
@cppforlife Updated this PR with what we've discussed. Is this what you had in mind?

release/release.go
```diff
@@ -28,26 +31,32 @@ type release struct {
 func NewRelease(
 	name string,
 	version string,
+	description string,
```
cppforlife: should we move all the strings into a mini struct ReleaseMeta so that it's easier to discern? positional args with same type really end up confusing.

petergtz:
@cppforlife sure, will make the changes.

petergtz:
@cppforlife did you have in mind using this struct everywhere then where these 3 fields appear, or just specifically in functions,  where the fact that parameters are not named make it kind of unreadable?

cppforlife:
> did you have in mind using this struct everywhere then where these 3 fields appear

i was thinking using it in places where these 3 fields appear. dont think should use that anywhere else since that would imply that extraneous info could be passed to functions.

release/release.go
petergtz: @cppforlife So I was about to make the requested changes, but after looking into it, I'm now wondering how useful they are. Here's why:
- `release.NewRelease` is the only candidate which uses positional arguments. All other uses are structs and therefore named. Agree that `release/release_test.go` is a bit ugly with all the `""`s, but don't thing it's too bad, because it's restricted mainly to this file.
- Other places like `release.Release` become a bit weird,  because the way name, version and description are treated is not the equivalent (description coming from the config, name and version coming from CLI args, see [`fs_release_dir.go`]()). That would require some weird constructs to make it work.
- `manifest.Manifest` models the yaml version of a manifest and I think it makes sense to maintain exactly the same structure as in the yaml file itself instead of using more nested levels in the Go struct than in the yaml.
- `director.Release` has yet again a slightly different structure: it contains an additional `VersionMark` function, and `Version` is of type `SemVer`, not `string`.

So, overall I don't think introducing a `ReleaseMeta` struct would actually help to make the code easier to read and maintain. What do you think?

jaresty:
merge story [#152973983](https://www.pivotaltracker.com/story/show/152973983)

jfmyers9:
Closing this issue due to inactivity.

This doesn't seem to be a priority. Feel free to open a new issue if you would like to discuss this feature further.

Best,

@jfmyers9

PR #300, opened 2017-08-21
allow leading slash for variables in manifest files while using vars-store file

MeBNoah:
Currently the BOSH CLI doesn't allow to reference variables with leading slashes, i.e. ((/foo)) when using a vars-store file. It is only allowed when using credhub instead of vars-file. In our use case we want to provide manifest files which can be deployed using credhub as well as using the vars-store file.

### DIFF
--- a/director/template/template.go
+++ b/director/template/template.go
@@ -87,7 +87,7 @@ func (t Template) interpolateRoot(obj interface{}, tracker varsTracker) (interfa
 type interpolator struct{}
 
 var (
-	interpolationRegex         = regexp.MustCompile(`\(\((!?[-\/\.\w\pL]+)\)\)`)
+	interpolationRegex         = regexp.MustCompile(`\(\((!?[-/\.\w\pL]+)\)\)`)
 	interpolationAnchoredRegex = regexp.MustCompile("\\A" + interpolationRegex.String() + "\\z")
 )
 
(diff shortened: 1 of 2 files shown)


director/template/template_test.go
```diff
@@ -21,6 +21,15 @@ var _ = Describe("Template", func() {
 		Expect(result).To(Equal([]byte("foo\n")))
 	})
 
+	It("can interpolate values with leading slash into a struct with byte slice", func() {
+		template := NewTemplate([]byte("((/key))"))
```
cppforlife: do you mind changing it to `/key/foo`?

MeBNoah: of course not ;) ... already done

director/template/template.go
```diff
@@ -87,7 +87,7 @@ func (t Template) interpolateRoot(obj interface{}, tracker varsTracker) (interfa
 type interpolator struct{}
 
 var (
-	interpolationRegex         = regexp.MustCompile(`\(\((!?[-\.\w\pL]+)\)\)`)
+	interpolationRegex         = regexp.MustCompile(`\(\((!?[-\/\.\w\pL]+)\)\)`)
```
dpb587-pivotal: Minor, but the forward slash doesn't need to be escaped. Maybe merger can change it if it isn't already changed.

PR #303, opened 2017-08-23
Helpful ssh error if multiple hosts/no hosts

drnic:
Example:

```
out/bosh ssh
Using environment 'https://10.58.111.49:25555' as client 'admin'

Using deployment 'consul'

Task 1054. Done

Running SSH:
  Interactive SSH only works for a single host at a time. Try `bosh ssh consul/6117487d-7277-42ef-adc0-da1e5e37afae`
```


cppforlife:
@drnic my main concern here is that a lot of these packages are meant to be used as libraries hence the error message is not true for them. somehow we need to propagate this info to cmd/ class.

PR #307, opened 2017-09-02
IPv6 fixes for ssh and create-env commands

cppforlife:


deployment/manifest/network.go
```diff
@@ -79,3 +74,15 @@ func (n Network) Interface(staticIPs []string, networkDefaults []NetworkDefault)
...
 }
+
+func ipMaskString(ipMask net.IPMask) string {
+	ip := net.IP(ipMask)
+
+	if p4 := ip.To4(); len(p4) == net.IPv4len {
+		return ip.String()
+	}
+
+	return fmt.Sprintf("%x:%x:%x:%x:%x:%x:%x:%x",
```
dpb587-pivotal: Is there a reason we need this custom encoding, rather than being able to rely on `ip.String()` to encode it?

ssh/session.go
```diff
@@ -153,7 +153,7 @@ func (r SessionImpl) makeKnownHostsFile() (boshsys.File, error) {
 
 	for _, host := range r.result.Hosts {
 		if len(host.HostPublicKey) > 0 {
-			content += fmt.Sprintf("%s %s\n", host.Host, host.HostPublicKey)
+			content += fmt.Sprintf("%s %s\n", printableHost{host}, host.HostPublicKey)
```
dpb587-pivotal: does ssh gateway support ipv6?

PR #309, opened 2017-09-06
Add `format` option to `repack-stemcell` command

Kiemes:
Format can be passed multiple times.

[#144851625](https://www.pivotaltracker.com/story/show/144851625)

Signed-off-by: Florian Nachtigall <florian.nachtigall@sap.com>


cdutra:
@Kiemes @tjvman Hey, we noticed this PR breaks the repack stemcell if no `--format` flag is passed and the stemcell does not already include `stemcell_formats`. It causes the director to error with the following error since it sees the field, but it's empty (possibly a separate director bug... but still don't think CLI should be adding an empty field)

    Task 4 | 23:57:24 | Error: stemcell_formats of this stemcell are not supported by available cpis

What would you think about not repacking stemcells with an empty `stemcell_formats`?

cppforlife:
@Kiemes yup, let's not add this field if it was not there before. may be
can use a pointer for that?

PR #312, opened 2017-09-08
Fetch deployment vms in parallel for vms command

knm3000:
[#145037925]( https://www.pivotaltracker.com/story/show/145037925)

Signed-off-by: Yulia Gaponenko <yulia.gaponenko1@de.ibm.com>


cmd/vms.go
```diff
@@ -55,12 +68,54 @@ func (c VMsCmd) printDeployments(instTable InstanceTable) error {
...
+	for i := 0; i < workSize; i++ {
+		errc := <-errorc
+		result := <-resultc
+		if errc != nil {
+			vmInfoErrors = append(vmInfoErrors, errc.Error())
+		}
+		vms[result.depName] = result.vmInfos
 	}
+	if len(vmInfoErrors) > 0 {
+		err := fmt.Errorf("%s", strings.Join(vmInfoErrors, "\n"))
```
tjvman: This seems like a good candidate for [`bosherr.MultiError`](https://github.com/cloudfoundry/bosh-utils/blob/master/errors/multi_error.go) - what do you think?

cmd/vms_test.go
```diff
@@ -35,6 +35,7 @@ var _ = Describe("VMsCmd", func() {
 
 		BeforeEach(func() {
 			opts = VMsOpts{}
+			opts.ParallelOpt = 5
```
tjvman: Is it possible to have additional test coverage for this new behavior? We see the integration test in the bosh director which looks good, but I wonder if we can cover at a unit level here.

Would it be practical to mock the `VMInfos` calls such that a first call is long, but second call is short? Then there could be some assertions around the second call being finished before the first to ensure the parallel behavior is working as expected. Just one idea - not sure if you have other ideas about it.

cmd/vms.go
```diff
@@ -55,12 +68,54 @@ func (c VMsCmd) printDeployments(instTable InstanceTable) error {
...
 
-func (c VMsCmd) printDeployment(dep boshdir.Deployment, instTable InstanceTable) error {
-	vmInfos, err := dep.VMInfos()
+type deploymentInfo struct {
+	depName string
+	vmInfos []boshdir.VMInfo
+}
+
+func parallelVMInfos(deployments []boshdir.Deployment, parallel int) (map[string][]boshdir.VMInfo, error) {
+	if parallel == 0 {
```
tjvman: This feels like surprising behavior - I would expect `--parallel=0` to mean "Don't parallelize this operation".  Could either set parallel to 1 if it's 0 or add a check in `printDeployments` that runs the normal `vmInfos` command if parallel is 0.

GarfieldIsAPhilosopher:
Merged in 146fdd744f08730fd9353781f5cedfb3fd195eda.

PR #319, opened 2017-09-20
Remove IPv6 bracketing from ssh shell out

cppforlife:
Still keep it for scp since it cannot determine if it's a IPv6 address or a file location.


dpb587-pivotal:
Merge Story: https://www.pivotaltracker.com/story/show/151540730

PR #332, opened 2017-10-14
Show colored diff in stdout during bosh deploy steps

aemengo:
screen shot 2017-10-14 at 10 03 43 am

Was originally present in the v1 cli

### DIFF
--- a/cmd/deploy_test.go
+++ b/cmd/deploy_test.go
@@ -226,8 +226,8 @@ releases:
 			Expect(err).ToNot(HaveOccurred())
 			Expect(deployment.DiffCallCount()).To(Equal(1))
 			Expect(ui.Said).To(ContainElement("  some line that stayed\n"))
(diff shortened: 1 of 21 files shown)


dpb587-pivotal:
Hi - thanks for taking a look at this! I think we'll need to consider a slightly different approach though. Rather than attempting to add methods for very specific use cases of add/remove lines, we probably need to plan out a way for generic coloring support which may involve some of the vendor'd libraries as well. That will require a bit more planning.

The coloring looks nice and has been mentioned on slack, too, but is it more for visual convenience or do you have another use case?

aemengo:
@dpb587-pivotal Thanks for getting back to me on this. This is more for visual convenience.

jaresty:
merge story [#152973883](https://www.pivotaltracker.com/story/show/152973883)

jfmyers9:

PR #343, opened 2017-10-26
Update bosh-gcscli

johnsonj:
New goodness in bosh-gcscli:
- Client falls back to `none` credentials mode when the application default credentials fail. This is to address behavior on GCE when the instance has a service account associated but does not have scope to the storage API. This prevented access to public buckets when attempting to fetch a token for authenticating.
- Client re-attempts "put" operation up to 3 times if it fails to upload
- `storage_class` settings are now only validated on `put`. This was causing failures when an actor had access to specific objects but not to the bucket settings (eg public objects)

bosh-cli change:
- client API has changed. There is now a single `New` method that expects a `config.GCSClient`.

Note:
- The `dep ensure -update github.com/cloudfoundry/bosh-gcscli` pulled in a ton of other package updates despite bosh-gcscli not updating anything else. Is that expected?

Validation:
- Used dev build to run the stackdriver-tools CI. This environment uses native GCS for the bosh release blobstore. No problems.
- Ran bin/test, following failures which appear unrelated:
```
UI AskForText [It] allows empty and non-empty text input 
/usr/local/google/home/jrjohnson/dev/src/github.com/cloudfoundry/bosh-cli/ui/ui_test.go:223

  Expected error:
      <errors.ComplexError>: {
          Err: {s: "Asking for text"},
          Cause: 0x19,
      }
      Asking for text: inappropriate ioctl for device
  not to have occurred
------------------------------
• Failure [0.000 seconds]
UI AskForPassword [It] allows empty and non-empty password input 
/usr/local/google/home/jrjohnson/dev/src/github.com/cloudfoundry/bosh-cli/ui/ui_test.go:246

  Expected error:
      <errors.ComplexError>: {
          Err: {
              s: "Asking for password",
          },
          Cause: 0x19,
      }
      Asking for password: inappropriate ioctl for device
  not to have occurred
```


dpb587-pivotal:
We've found that `dep ensure -update *` doesn't actually limit the update to the package argument, which is why all the other packages were updated.

Can you run `dep prune` as part of the `dep ensure` and squash the result? This will avoid committing the 3,800,000 other lines of code that we don't rely on. Keeps the diff for review a bit smaller, too :)

Thanks for calling out the test failures - it does look like they're related to one of the other side effect bumps that we'll have to look into.

johnsonj:
Done! that looks a lot better.

I'm guessing the test failure is related to changes in `github.com/vito/go-interact`. It doesn't repro on `develop`

johnsonj:
@dpb587-pivotal anything I can do to help get this merged? 

I can take a shot at updating the dep without updating `github.com/vito/go-interact`

cppforlife:
@johnsonj we are planning to do bunch of prs beginning of the next week. we'll pull this in shortly.

johnsonj:
Thanks for the update @cppforlife ! I'll be on call if you need me :)

jaresty:
merge story [#152973820](https://www.pivotaltracker.com/story/show/152973820)

johnsonj:
Thank you!

PR #351, opened 2017-11-03
Add --cloud-properties to vms

manno:
Show actual cloud properties of every VM

https://www.pivotaltracker.com/story/show/151157765

### DIFF
--- a/cmd/instance_table.go
+++ b/cmd/instance_table.go
@@ -18,14 +18,15 @@ type InstanceTableValues struct {
 	IPs          boshtbl.Value
 
 	// Details
-	VMCID        boshtbl.Value
-	DiskCIDs     boshtbl.Value
-	AgentID      boshtbl.Value
-	Index        boshtbl.Value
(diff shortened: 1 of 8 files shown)


dpb587-pivotal:
Hi - thanks, this sounds like a nice addition. Few questions...

 * Can you provide an integration test for this as well?
 * Does this work for both vm_resources and traditional vm_type+vm_extensions?
 * Is the cloud properties output going to be one key-value per line like the events context? Looks like yes.

friegger:
@dpb587-pivotal 
- we'll provide an integration test in a commit to bosh, that we push and reference here after we merged the PR
- yes and the integration tests will cover both cases
- yes, it's yaml, so it will be one key-value per line, since cloud properties is a hash with just keys and string values

friegger:
Integration tests: https://github.com/cloudfoundry/bosh/commit/a1cb7aa5f00512f52fdf3e41a3ae942cea07bca3

PR #361, opened 2017-11-27
Add bash completion with default handling

### LINKED ISSUE #175: bring back autocomplete - without breaking anything else :)
Commit https://github.com/cloudfoundry/bosh-cli/commit/2c59907fe986e0a38837b5847648bf30aebbb6ee has just removed autocomplete for CLI commands, because completion for files was broken (see https://github.com/cloudfoundry/homebrew-tap/issues/28).

But in the meantime I learned to appreciate the beauty of autocomplete for commands. Can we get that back, but now for realz and without breaking anything else? Thanks :)

friegger:
Re-adds bash completion with `default` option as suggested in https://github.com/cloudfoundry/homebrew-tap/issues/28.

In case there is no completion implemented for custom go-flags args or option, it falls back to default file and folder completion. As a small drawback, this also happens where file completion might not be appropriate, as for example with the `-d` option.

Tested on Mac and on Ubuntu, probably Mac is just fine as this is for homebrew.  

Fixes #175

### DIFF
--- a/ci/tasks/update-homebrew-formula.sh
+++ b/ci/tasks/update-homebrew-formula.sh
@@ -22,6 +22,19 @@ class BoshCli < Formula
   def install
     binary_name = build.with?("bosh2") ? "bosh2" : "bosh"
     bin.install "bosh-cli-#{version}-darwin-amd64" => binary_name
+    (bash_completion/"bosh-cli").write <<-completion
+      _#{binary_name}() {
+          # All arguments except the first one
+          args=("\${COMP_WORDS[@]:1:\$COMP_CWORD}")
+          # Only split on newlines
+          local IFS=$'\n'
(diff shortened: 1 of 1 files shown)


tjvman:
Merge story: #153676868

friegger:
@tjvman Sure! Thanks.

voelzmo:
*ping* @cppforlife @dpb587-pivotal @tjvman can we pull this in? Any objections to merging this?

cppforlife:
didnt we break something last time auto conpletion was merged?

voelzmo:
IIRC file completion didn't work anymore. I think this was specifically tested and fixed by this new approach.

dpb587-pivotal:
Looks good and works for me.

Issue #20, opened 2016-10-10
`create-release` gives a confusing error message if run from non-release directory


cunnie:
## Steps to reproduce

Run `bosh create-release` from any non-release directory, e.g. `/`:

``` bash
 2016-10-10 09:33:07 ☆ ruby 2.0.0p648 (system) moncada in /
○ → bosh create-release
Expected non-empty 'final_name' in config '/config/final.yml'

Exit code 1
```

The output from the Ruby CLI was much more helpful:

``` bash
 2016-10-10 09:35:57 ☆ ruby 2.2.5p319 moncada in /
○ → bosh-ruby create release
Sorry, your current directory doesn't look like release directory
```

Zanadar:
I think this can be closed, as it was addressed in https://github.com/cloudfoundry/bosh-cli/pull/368

cunnie:
Thanks @Zanadar !, closing.

PR #368, opened 2017-12-11
Warn when neither public nor private release config directories exist

Zanadar:
I took a crack at addressing https://github.com/cloudfoundry/bosh-cli/issues/20

### DIFF
--- a/releasedir/fs_config.go
+++ b/releasedir/fs_config.go
@@ -1,6 +1,8 @@
 package releasedir
 
 import (
+	"strings"
+
 	bosherr "github.com/cloudfoundry/bosh-utils/errors"
 	boshsys "github.com/cloudfoundry/bosh-utils/system"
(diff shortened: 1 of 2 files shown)


tjvman:
Merge story: [#153676830](https://www.pivotaltracker.com/story/show/153676830)

Zanadar:
@tjvman Will do in the future, thanks!  Anything I can do to move this one forward, or are we good for now?

releasedir/fs_config.go
```diff
@@ -69,7 +71,7 @@ func (c FSConfig) Name() (string, error) {
 
 func (c FSConfig) SaveName(name string) error {
 	publicSchema, _, err := c.read()
-	if err != nil {
+	if err != nil && !strings.Contains(err.Error(), "isn't a release directory") {
```
cppforlife: shouldnt this always be `if err != nil {`

PR #369, opened 2017-12-14
Asynchronous Update and Delete Deployments

terminatingcode:
We are using the bosh-cli as a library on the [Services-Enablement](https://github.com/pivotal-cf/on-demand-service-broker) team and were encountering issues with the Cloud Controller timing out due to the synchronous nature of the Update and Delete Deployment functions. We return the task id of the tasks so that polling may be done on our side.

We added new methods in order to preserve the behaviour of the existing functions and to not introduce any breaking changes. Please let us know if that's appropriate.

Any feedback welcome.

Thanks

with @kirederik

### DIFF
--- a/director/deployment.go
+++ b/director/deployment.go
@@ -170,11 +170,18 @@ func (d DeploymentImpl) ExportRelease(release ReleaseSlug, os OSVersionSlug, job
 }
 
 func (d DeploymentImpl) Update(manifest []byte, opts UpdateOpts) error {
+	opts.async = false
+	_, err := d.client.UpdateDeployment(manifest, opts)
+	return err
+}
+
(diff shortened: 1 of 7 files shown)


tjvman:
Merge story: [#153676801](https://www.pivotaltracker.com/story/show/153676801)

terminatingcode:
The services-enablement team have pursued an alternative route to resolve our issue. Instead of requiring the bosh-cli to respond asynchronously, we make the call in a go-routine and handle our responses to other APIs asynchronously (examples of which can be found [here](https://github.com/pivotal-cf/on-demand-service-broker/tree/master/boshdirector))

cppforlife:
that sounds like a nicer alternative. thank you.

Issue #375, opened 2017-12-22
Assumes director and uaa share same ca cert


jmcarp:
The cli assumes that the director cert and uaa cert are signed by the same ca. If that's not the case, and the user sets `BOSH_CA_CERT` to the director ca cert, it's impossible to authenticate via uaa. It would be useful to allow different ca certs for the director and uaa--we could add an optional `BOSH_UAA_CA_CERT` option and default to the director ca cert if not provided.

@LinuxBozo

dpb587-pivotal:
For what it's worth, the CA cert passed to BOSH can contain multiple, independent certificates. The CLI then trusts certs based from any of the certificates passed in the CA cert data. That's one alternative to trusting system certs or requiring the same CA for both.

PR #376, opened 2017-12-22
Always trust system certificates.

jmcarp:
Because the cli [assumes the same ca certificate for the director and the uaa](https://github.com/cloudfoundry/bosh-cli/issues/375), authentication fails when the director cert is signed by a private ca and the uaa cert is signed by a public ca. In this case, users *could* append the public ca to the private ca and pass both to `BOSH_CA_CERT`, but we shouldn't have to duplicate certs already trusted by the system. This patch handles this use case without extra cert copy-pasta by starting from the system certs and appending the director ca cert if specified.

@LinuxBozo

### DIFF
--- a/director/factory_config.go
+++ b/director/factory_config.go
@@ -7,7 +7,6 @@ import (
 	"strconv"
 	"strings"
 
-	"github.com/cloudfoundry/bosh-utils/crypto"
 	bosherr "github.com/cloudfoundry/bosh-utils/errors"
 )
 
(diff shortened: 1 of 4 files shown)


LinuxBozo:
This is especially useful when you want to expose UAA publicly for authentication of other tools (more than just BOSH), and have the BOSH director remain private.

aeijdenberg:
@jmcarp - a pattern I've seen elsewhere is to use the system CA certs by default, but if a list of CA certs is explicitly specified, then use those, and those only instead.

I would seem surprising to me if I specified a CA cert, but then certs not signed by that CA (e.g. by chaining to a public trusted root) were also accepted.

I realise that wouldn't solve the issue that sparked this PR though. Maybe it would be better to go with your suggestion in the linked issue, that is make the CA for UAA a separately configured option? (and for either config, if left empty, default to the system CA pool?)

cppforlife:
> a pattern I've seen elsewhere is to use the system CA certs by default, but if a list of CA certs is explicitly specified, then use those, and those only instead.

i agree with the above.

i could see adding an additional flag such as `--ca-cert-system` (`BOSH_CA_CERT_SYSTEM=true`) for this use case though we would have to save that boolean into .bosh/config.

cppforlife:
@jmcarp are you interested in adding `--ca-cert-system` ?

jfmyers9:

PR #377, opened 2018-01-02
Use filepath so paths work on Windows

Fydon:


cppforlife:
merged via https://github.com/cloudfoundry/bosh-cli/commit/4dfb548e06d75d16cd8d9d9b54238b09620c1064

PR #382, opened 2018-01-04
Enable ERB trim support

### LINKED ISSUE #381: Bosh create-env fails when release template uses `-%>`
This difference in behaviour can be trace back to a different ERB uses in the evaluation contexts:
```ruby
# bosh create-env
erb = ERB.new(File.read(src_path))

# director
template = ERB.new(File.read(template_name), safe_level = nil, trim_mode = "-")
```
source [cli](https://github.com/cloudfoundry/bosh-cli/blob/master/templatescompiler/erbrenderer/template_evaluation_context_rb.go#L177), [director](https://github.com/cloudfoundry/bosh/blob/master/src/bosh-template/lib/bosh/template/renderer.rb#L15)

Is this difference in behaviour intended or would a PR to fix it be merged?

rkoster:
Fixes #381

### DIFF
--- a/templatescompiler/erbrenderer/template_evaluation_context_rb.go
+++ b/templatescompiler/erbrenderer/template_evaluation_context_rb.go
@@ -72,7 +72,7 @@ class TemplateEvaluationContext
     yield *values
     InactiveElseBlock.new
   end
-  
+
   def if_link(name)
     false
   end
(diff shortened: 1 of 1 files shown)


voelzmo:
Thanks @rkoster!
Could you maybe also add a test similar to what I did in https://github.com/cloudfoundry/bosh/pull/1482?

cppforlife:
@voelzmo might not be worth it for this specific case.

PR #383, opened 2018-01-05
Fix file descriptor leak

aeijdenberg:
Response bodies must always be closed by someone, and they were being ignored in this case.

If someone wants a bigger project, I think this entire `AdjustableClient` class needs a rethink. As I understand it, an `http.Request` carries some state, so it probably isn't safe to just retry it.

Perhaps it would be better to check if the OAuth token isn't expired before sending it to begin with?

Anyway for now I wanted to keep this patch small, as I believe it will fix <https://github.com/bosh-prometheus/bosh_exporter/issues/14> which is affecting us in production.

### DIFF
--- a/director/adjustable_client.go
+++ b/director/adjustable_client.go
@@ -41,6 +41,7 @@ func (c AdjustableClient) Do(req *http.Request) (*http.Response, error) {
 	}
 
 	if c.adjustment.NeedsReadjustment(resp) {
+		resp.Body.Close()
 		err := c.adjustment.Adjust(req, true)
 		if err != nil {
 			return nil, err
--- a/director/adjustable_client_test.go
+++ b/director/adjustable_client_test.go
@@ -8,12 +8,42 @@ import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 
-	. "github.com/cloudfoundry/bosh-cli/director"
-	fakedir "github.com/cloudfoundry/bosh-cli/director/directorfakes"
 	"io"
 	"io/ioutil"
 	"strings"
+
+	. "github.com/cloudfoundry/bosh-cli/director"
+	fakedir "github.com/cloudfoundry/bosh-cli/director/directorfakes"
 )
+
+type FakeResponseBodyFactory struct {
+	Bodies []*FakeResponseBody
+}
+
+func (frbf *FakeResponseBodyFactory) Verify() {
+	for _, frb := range frbf.Bodies {
+		Expect(frb.ClosedCount).To(Equal(1))
+	}
+}
+
+func (frbf *FakeResponseBodyFactory) NewResponseBody() io.ReadCloser {
+	rv := &FakeResponseBody{}
+	frbf.Bodies = append(frbf.Bodies, rv)
+	return rv
+}
+
+type FakeResponseBody struct {
+	ClosedCount int
+}
+
+func (frb *FakeResponseBody) Read(p []byte) (n int, err error) {
+	return 0, nil
+}
+
+func (frb *FakeResponseBody) Close() error {
+	frb.ClosedCount += 1
+	return nil
+}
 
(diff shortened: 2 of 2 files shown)


aeijdenberg:
Gentle ping - would be great if someone could take a quick look and merge this (it's just a one-liner to `Close()` a file descriptor), as the hacky workaround we have in prod (daily cron to restart the job) isn't great, so I'd like to get the `bosh_exporter` updated to use this code. Thanks!

cppforlife:
@aeijdenberg any chance you could add a unit test to director/adjustable_client.go ?

aeijdenberg:
Sure thing @cppforlife , 87 lines of unit tests are added for the 1 line fix. :)

(turns out my one-liner did actually break the tests that were there, so I fixed them up)

The more I look at this bit of code, I'm pretty sure there's some improvements we can make (I've done UAA token refreshing in a number of other apps), but I'll keep this PR restricted to fixing the file descriptor leak.

aeijdenberg:
Gentle ping again - I added tests as requested last week.

aeijdenberg:
Ping again. This is a one line change to fix a simple resource leak. Would appreciate a review and merge.

The [official Go `http.Client` docs](https://golang.org/pkg/net/http/#Client.Do) are quite on the behavior that needs to occur, and the current code simply discards the response without calling `Close()`:

> If the returned error is nil, the Response will contain a non-nil Body which the user is expected to close.

dpb587-pivotal:
Units and integrations are green.

director/adjustable_client.go
```diff
@@ -41,6 +41,7 @@ func (c AdjustableClient) Do(req *http.Request) (*http.Response, error) {
 	}
 
 	if c.adjustment.NeedsReadjustment(resp) {
+		resp.Body.Close()
```
dpb587-pivotal: @aeijdenberg Can you help me understand why this fixes what you're seeing? I think this `Close` close the request body file descriptor. Which would mean subsequent requests will most definitely fail if it has a body and is trying to reuse this? I'd expect a `file already closed` type of error on retry.

aeijdenberg: @dpb587-pivotal - thanks for looking at this PR. Once we get inside of this `if` statement, the value `resp` is no longer used (I agree this function as written is quite confusing).

i.e. reaching this point, the function will either return on line 47, or line 52. In either case, `resp` is not accessed, and thus is never closed.

dpb587-pivotal: Ah, I read too quickly. This is the response body, not the request body. Thanks.

aeijdenberg:
Ping again. We had another production issue on another system last week due this this bug as a root cause. Are there any unaddressed concerns preventing merging? (it's a one line fix)

Thanks for your help.

dpb587-pivotal:
@aeijdenberg have you tested bosh_exporter/production with this changeset to verify it fixes the issue happening there?

aeijdenberg:
@dpb587-pivotal we've verified that:

1. this code produces a file descriptor leak
2. that this fix stops this descriptor leak
2. that this code is called by the bosh_exporter
3. that the bosh_exporter suffers from a file descriptor leak.

If this code is indicative, I wouldn't be surprised if there are additional sources of file descriptor leaks, but we need to start somewhere, and I can't recall whether I spun up a local bosh_exporter to see if it was positively affected by this change - I think I did, but we certainly didn't go to the effort of forking a set of bosh releases to test in prod to verify this fixes all the issues.

The Go standard library docs are very clear on the expected client behavior, which prior to this patch the client is not doing:

`If the returned error is nil, the Response will contain a non-nil Body which the user is expected to close` (<https://golang.org/pkg/net/http/#Client.Do>)

dpb587-pivotal:
Merged - thanks for sending the pull request (and sorry this turned out to be a pain).

aeijdenberg:
Thanks @dpb587-pivotal  :)

Issue #329, opened 2017-10-12
can the director client library be in a separate repo than bosh-cli codebase


zankich:
I want to use the bosh director client library in my code and it's unfortunate that I have to pull down all the dependencies for the cli.

If I use `dep` to pull down `bosh-cli` as a dependency, it will not copy the inner `vendor` directory, so I end up having to deal with issues like this:

```
# github.com/cloudfoundry-incubator/cf-mysql-acceptance-tests/vendor/github.com/cloudfoundry/bosh-cli/ui
vendor/github.com/cloudfoundry/bosh-cli/ui/file_reporter.go:32:42: cannot use r.buildBar(size) (type *pb.ProgressBar) as type io.Writer in argument to io.MultiWriter:
	*pb.ProgressBar does not implement io.Writer (missing Write method)
		have pb.write()
		want Write([]byte) (int, error)
vendor/github.com/cloudfoundry/bosh-cli/ui/file_reporter.go:36:9: undefined: pb.New
vendor/github.com/cloudfoundry/bosh-cli/ui/file_reporter.go:44:15: undefined: pb.U_BYTES
```

This requires me to have to manually lock dependencies to the versions that the `bosh-cli` wants, which means that I constantly have to make sure my dependencies match what the `bosh-cli` mainline wants.

It would be awesome if the bosh client library had very minimal dependencies so that app developers who would like to programmatically integrate with bosh don't need to pull down the universe.

kieron-pivotal:
We've just come across this problem as well. Would it be a solution to include 
```
[[constraint]]
name = "github.com/cheggaaa/pb"
version = "=1.0.18"
```
in Gopkg.toml so that it resolves correctly as a transitive dependency?

jfmyers9:
If anyone is looking for an example on how to successfully do this, [this is a great example](https://github.com/cloudfoundry/bosh-deployment-resource/blob/24b8989c59e846b0b3afd2816469ea39afbc868f/Gopkg.toml).

Also, since this is the preferred library for programatic access to the bosh director, are there any plans to improve this user experience any time soon?

jfmyers9:
With recent changes, `go mod` and other vendoring tools should be able to successfully vendor the director library. Due to this, I'm going to close this issue. Please feel free to open a new issue if you are still having problems.

PR #391, opened 2018-01-28
director package doesn't depend on ui package

rosenhouse:
I'm trying to use the `director` package in another program.

I was surprised to find that `director` depends on `ui`.  But that's only so that it can use a `ReadSeekCloser` type.

This commit breaks that dependency.

I think it may help with https://github.com/cloudfoundry/bosh-cli/issues/329

package import graphs:
- [current master](https://godoc.org/github.com/cloudfoundry/bosh-cli/director?import-graph&hide=2)
- [with this change](https://godoc.org/github.com/rosenhouse/bosh-cli/director?import-graph&hide=2)

### DIFF
--- a/director/interfaces.go
+++ b/director/interfaces.go
@@ -5,7 +5,7 @@ import (
 	"os"
 	"time"
 
-	"github.com/cloudfoundry/bosh-cli/ui"
+	bio "github.com/cloudfoundry/bosh-cli/io"
 	semver "github.com/cppforlife/go-semi-semantic/version"
 )
 
@@ -100,7 +100,7 @@ type StemcellArchive interface {
 //go:generate counterfeiter . FileReporter
 
 type FileReporter interface {
-	TrackUpload(int64, io.ReadCloser) ui.ReadSeekCloser
+	TrackUpload(int64, io.ReadCloser) bio.ReadSeekCloser
 	TrackDownload(int64, io.Writer) io.Writer
 }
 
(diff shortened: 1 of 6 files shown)


dpb587-pivotal:
I think this makes a lot of sense. Will discuss more with team tomorrow.

evanfarrar:
👍  We tested this PR out for part of [bosh-test](https://github.com/pivotal-cf-experimental/bosh-test) and it worked for us. We also verified that a CLI built with this still worked.

director/interfaces.go
```diff
@@ -95,7 +94,10 @@ type StemcellArchive interface {
 //go:generate counterfeiter . FileReporter
 
 type FileReporter interface {
-	TrackUpload(int64, io.ReadCloser) ui.ReadSeekCloser
```
cppforlife: let's make type ReadSeekCloser interface in director package. too ugly with so many same anonymous interfaces.

rosenhouse: Done

rosenhouse:
@cppforlife @dpb587 I've updated the PR, the `ReadSeekCloser` type now lives in the `director` package.

So now `ui` depends on `director`.  Which is still better than the reverse.

cppforlife:
> So now ui depends on director. Which is still better than the reverse.

i meant we should just have two same interface decl in both pkgs so that they dont depend on each other.

rosenhouse:
That is not trivial, because go doesn’t have covariant interfaces.   Try it
yourself :-)

cppforlife:
@rosenhouse fell off the truck, sorry. you are right about covariant interfaces, forgot. let's add `io` package with just that interface and use it in director and ui packages.

rosenhouse:
@cppforlife finally got around to this.  there's now a dedicated `io` package, used by `director` and `ui`.

PR #404, opened 2018-03-20
VarsStore accepts file containing only yaml header

XenoPhex:
Previously, if the vars-store file only contained `---`, then the CLI
would panic when trying to set a variable on the map created from the
unmarshaled file.

Signed-off-by: Anand Gaitonde <agaitonde@pivotal.io>

### DIFF
--- a/cmd/vars_fs_store.go
+++ b/cmd/vars_fs_store.go
@@ -97,6 +97,9 @@ func (s VarsFSStore) load() (boshtpl.StaticVariables, error) {
 			return vars, bosherr.WrapErrorf(err, "Deserializing variables file store '%s'", s.path)
 		}
 	}
+	if vars == nil {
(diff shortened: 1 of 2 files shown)


XenoPhex:
Related issue: https://github.com/go-yaml/yaml/issues/347

dpb587-pivotal:
Interestingly, it seems [empty nodes](http://www.yaml.org/spec/1.2/spec.html#id2786563) are supposed to be empty scalars whose value is not explicitly defined. I don't think, technically, a `map` qualifies as a scalar; otherwise we could try to suggest an empty node should not touch the go internal value at all (allowing it retain whatever it's native empty value is; which would be good in our case).

 > YAML allows the node content to be omitted in many cases. Nodes with empty content are interpreted as if they were plain scalars with an empty value. Such nodes are commonly resolved to a “null” value.

Issue #392, opened 2018-02-05
"Persistent disk with volume id 'xxx' could not be found" when creating bosh director


gossion:
I got an error when creating bosh env on Azure, this is the second time I got this issue, it is not always reproduced. The error log is as below, which is quite similar to issue #110 (already fixed), but this should be a different issue.

### Error when deploying bosh director:  
``
[CLI] 2018/02/03 13:05:37 ERROR - Deploying: Creating instance 'bosh/0': Updating instance disks: Updating disks: Deploying disk: Mounting disk: Sending 'get_task' to the agent: Agent responded with error: Action Failed get_task: Task 0dd59011-9dfa-45c6-6f16-b6c046d87dfe result: Persistent disk with volume id 'caching:None;disk_name:bosh-disk-data-bae04939-b5dd-4330-b36e-8cc1e65505ef;resource_group_name:ICA-RG-ManagedCF-2-3-4-38-57' could not be found
``
see more detail log in this [link](https://raw.githubusercontent.com/gossion/logs/master/run.log)

I tried to get bosh agent log of the director VM, and found that in `/var/vcap/bosh/settings.json` the persistent disk is not listed in `disks.persistent`, but I am sure that the data disk was already attached to the VM.
```json
  "disks": {
    "system": "/dev/sda",
    "ephemeral": {
      "host_device_id": "{f8b3781b-1e82-4818-a1c3-63d806ec15bb}",
      "lun": "0",
      "path": "/dev/sdc"
    },
    "persistent": {},
    "raw_ephemeral": null
  },
```

### Error got from bosh agent log:
```
2018-02-03_13:05:25.64117 [DelayedAuditLogger] 2018/02/03 13:05:25 DEBUG - Logging CEF:0|CloudFoundry|BOSH|1|agent_api|/agent|1|duser=mbus requestMethod=POST src=52.166.34.200 spt=56098 shost=2da16db4-c236-486a-4e68-3a3c67b61e08 cs1=HOST=&X_REAL_IP=&X_FORWARDED_FOR=&X_FORWARDED_PROTO=&USER_AGENT= cs1Label=httpHeaders cs2=basic cs2Label=authType cs3=200 cs3Label=responseStatus  to syslog
2018-02-03_13:05:26.48262 [attemptRetryStrategy] 2018/02/03 13:05:26 DEBUG - Making attempt #9 for *httpclient.RequestRetryable
[...]

PR #405, opened 2018-03-22
log sshtunnel errors

gossion:
This PR is to improve logging around ssh tunnel errors. With this PR, it will be easier to debug issues like #392 in future.

### DIFF
--- a/deployment/instance/instance.go
+++ b/deployment/instance/instance.go
@@ -100,6 +100,13 @@ func (i *instance) WaitUntilReady(
 			sshTunnel := i.sshTunnelFactory.NewSSHTunnel(sshTunnelOptions)
 			go sshTunnel.Start(sshReadyErrCh, sshErrCh)
 
+			go func() {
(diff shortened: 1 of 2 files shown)


dpb587-pivotal:
🔗 Pivotal Tracker, [Story #156214997](https://www.pivotaltracker.com/story/show/156214997)

deployment/instance/instance.go
```diff
@@ -100,6 +100,13 @@ func (i *instance) WaitUntilReady(
 			sshTunnel := i.sshTunnelFactory.NewSSHTunnel(sshTunnelOptions)
 			go sshTunnel.Start(sshReadyErrCh, sshErrCh)
 
+			go func() {
+				for {
```
jfmyers9: Currently this for loop will never terminate, and will hard loop once the `sshErrCh` is closed.

This can be improved by using:
```
for err := range sshErCh {
  ...
}
```

PR #408, opened 2018-03-29
add curl command for debugging

cppforlife:
adds curl command that is pre-pointed at the Director. it will insert authentication headers.

```
$ bosh curl /deploments | jq .
$ bosh curl -X POST /configs --body <(echo "body")
$ bosh curl -X GET /deploments -H "Useless: header" --show-headers
```


cmd/curl.go
```diff
@@ -0,0 +1,93 @@
...
+	if len(pieces) != 2 {
+		return bosherr.Errorf("Expected header '%s' to be in format 'name: value'", data)
+	}
+
+	if len(pieces[0]) == 0 {
+		return bosherr.Errorf("Expected header '%s' to specify non-empty name", data)
+	}
+
+	if len(pieces[1]) == 0 {
+		return bosherr.Errorf("Expected header '%s' to specify non-empty value", data)
```
dpb587-pivotal: Empty header values are allowed in HTTP.

dpb587-pivotal:
What is the specific value and use cases for this?

drnic:
Will it return nice concise JSON? Or just return the pseudo JSON that the cli current returns with the —json flag? 
		
		

		
	




On Fri, Mar 30, 2018 at 6:33 AM +1000, "Dmitriy Kalinin" <notifications@github.com> wrote:













What is the specific value and use cases for this?


helps debugging responses from the director
helps me in acceptance
helps working with new apis like (POST /links) that are not planned to exposed thru a command

cppforlife:
> Will it return nice concise JSON? Or just return the pseudo JSON that the
cli current returns with the —json flag?

it will return whatever response director returns, in some cases it's not
even json.

drnic:
```
out/bosh curl /deployments
out/bosh curl /deployments/cf | jq -r .manifest
```
This is great.

I was looking at upgrading https://github.com/cloudfoundry-community/stannis (which uses a very very old gogobosh version) and discovered that new gogobosh and bosh-cli's own example.go are still quite tedious to get started with (required explicit UAA setup; don't look for config from env vars etc). So shelling out to `bosh curl` might be super nice way to abstract away all the configuration of which BOSH by default.

drnic:
Bump

jraqula:
thump

pivotal-jamil-shamy:
Merged into develop

PR #409, opened 2018-03-30
Add unalias-env command

### LINKED ISSUE #284: Bosh CLI command that can remove an environment alias?
We are working on teardown scripts and `delete-env` does not remove the environment alias. Is there an argument that can be provided to `bosh alias-env` to remove a previously created alias? Or maybe there could be a new `delete-alias` command?

For now we are forced to manually clean up our `~/.bosh/config` but since we have multiple alias its an ugly shell script.

zmb3:
Fixes #284

### DIFF
--- a/cmd/config/fs_config.go
+++ b/cmd/config/fs_config.go
@@ -98,7 +98,7 @@ func (c FSConfig) AliasEnvironment(url, alias, caCert string) (Config, error) {
 	return config, nil
 }
 
-func (c FSConfig) DeleteAlias(alias string) (Config, error) {
+func (c FSConfig) UnaliasEnvironment(alias string) (Config, error) {
 	if len(alias) == 0 {
 		return nil, bosherr.Error("expected non-empty environment alias")
 	}
--- a/cmd/config/fs_config.go
+++ b/cmd/config/fs_config.go
@@ -98,6 +98,26 @@ func (c FSConfig) AliasEnvironment(url, alias, caCert string) (Config, error) {
 	return config, nil
 }
 
+func (c FSConfig) DeleteAlias(alias string) (Config, error) {
+	if len(alias) == 0 {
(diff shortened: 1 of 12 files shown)


cppforlife:
@zmb3 should we name this command `unalias-env`?

zmb3:
@cppforlife was actually going to do that, but Danny suggested delete-alias in the linked issue so I stuck with that. No strong opinions here.

cppforlife:
will check in with danny tomorrow...

dpb587-pivotal:
Grammatically/verbiage-wise... `delete-alias` makes more sense to me than `unalias-env`. But I do see the inconsistency between `alias-env` and `delete-alias`. Not sure there's a great choice, and I don't have a strong opinion either way.

cppforlife:
@zmb3 all right, let's go with `unalias-env`.

zmb3:
@cppforlife @dpb587-pivotal should be all set.

Despite the grammar I actually do like `unalias-env` - it sounds less destructive than `delete-alias`.

cmd/config/fs_config.go
```diff
@@ -98,6 +98,26 @@ func (c FSConfig) AliasEnvironment(url, alias, caCert string) (Config, error) {
 	return config, nil
 }
 
+func (c FSConfig) DeleteAlias(alias string) (Config, error) {
```
cppforlife: being picky... can we rename this to UnaliasEnvironment to be consistent with AliasEnvironment.

zmb3: Sure.  I only left this one because from the perspective of the config we are actually _deleting_ the alias.

zmb3:
friendly ping @cppforlife @dpb587-pivotal

mfine30:
closing and re-opening so we can prioritize

Issue #334, opened 2017-10-16
--column panics when passed a header that doesn't exist


ishustava:
```
$ bosh deployments --column=Release
```
fails with: 
```
[CLI] 2017/10/16 11:42:59 ERROR - Panic: Failed to find header: release
********************
goroutine 1 [running]:
runtime/debug.Stack(0x98b8a0, 0xc4203371c0, 0x1e)
	/usr/local/go/src/runtime/debug/stack.go:24 +0x79
main.handlePanic()
	/tmp/build/a9832f70/gopath/src/github.com/cloudfoundry/bosh-cli/main.go:110 +0x224
panic(0x8dbf00, 0xc420323300)
	/usr/local/go/src/runtime/panic.go:458 +0x243
github.com/cloudfoundry/bosh-cli/cmd.Cmd.Execute.func1(0xc42026c348)
	/tmp/build/a9832f70/gopath/src/github.com/cloudfoundry/bosh-cli/cmd/cmd.go:47 +0xe7
panic(0x8dbf00, 0xc420323300)
	/usr/local/go/src/runtime/panic.go:458 +0x243
github.com/cloudfoundry/bosh-cli/ui.(*ConfUI).PrintTable(0xc42007fe00, 0x0, 0x0, 0xa1c75f, 0xb, 0xc42020f860, 0x5, 0x5, 0xc420322d10, 0x1, ...)
	/tmp/build/a9832f70/gopath/src/github.com/cloudfoundry/bosh-cli/ui/conf_ui.go:87 +0xf2
github.com/cloudfoundry/bosh-cli/cmd.DeploymentsTable.Print(0xc4200aab00, 0x7, 0x8, 0xf07e60, 0xc42007fe00, 0x0, 0xf584c8)
	/tmp/build/a9832f70/gopath/src/github.com/cloudfoundry/bosh-cli/cmd/deployments_table.go:61 +0x924
github.com/cloudfoundry/bosh-cli/cmd.DeploymentsCmd.Run(0xf07e60, 0xc42007fe00, 0xf0b0c0, 0xc420018630, 0x2e, 0xf09bc0)
	/tmp/build/a9832f70/gopath/src/github.com/cloudfoundry/bosh-cli/cmd/deployments.go:23 +0xd8
[...]

PR #358, opened 2017-11-08
Fixes #334. WIP DON'T MERGE.

sahilm:
Hello -

This PR is an attempt to fix #334. The output of `bosh deployments --column Release` is now:

```
$ bosh deployments --column Release
Using environment 'https://10.0.0.6:25555' as client 'admin'

Failed to find header: release

Exit code 1
```

I've hacked together a solution and would like some feedback before I proceed to add tests etc. Is this the right approach to solve this problem? Any other feedback is welcome.

Thanks.

PR #415, opened 2018-04-13
Change UI PrintTable to return error rather than panic when column header missing

terminatingcode:
This is mostly a duplicate of the stale PR #358 

Fixes issue #334 

Feel free to accept the other PR and close this if preferable.


Amit-PivotalLabs:
+1, I just hit this issue.

PR #421, opened 2018-05-03
'bosh logs' support for Windows

### LINKED ISSUE #419: 'bosh logs' command failure on Windows
The `bosh logs` command fails on Windows because the CLI attempts to move the temp file whilst it still has an open handle, and Windows doesn't like that too much:

```
λ bosh -e vbox -d zookeeper logs zookeeper/1
Using environment '192.168.50.6' as client 'admin'

Using deployment 'zookeeper'

Task 11

Task 11 | 20:16:20 | Fetching logs for zookeeper/0e82df1a-35e7-49cb-8f2d-de99115c69cd (1): Finding and packing log files (00:00:01)

Task 11 Started  Thu May  3 20:16:20 UTC 2018
Task 11 Finished Thu May  3 20:16:21 UTC 2018
Task 11 Duration 00:00:01
Task 11 done

Downloading resource 'd9f61878-320c-49bb-4b52-14940388b197' to 'C:\Users\rcw5\zookeeper.zookeeper.1-20180503-211621-7925707.tgz'...

                                                                           0.00%
Downloading logs:
  Moving to final destination:
    rename C:\Users\rcw5\.bosh\tmp\director-resource-d9f61878-320c-49bb-4b52-14940388b197661277999 C:\Users\rcw5\zookeeper.zookeeper.1-20180503-211621-7925707.tgz: The process cannot access the file because it is being used by another process.

Exit code 1
```

Everything works just fine if the file is closed before the move takes place:

```
λ bosh2 -e vbox -d zookeeper logs zookeeper/1
Using environment '192.168.50.6' as client 'admin'

Using deployment 'zookeeper'

Task 14

Task 14 | 20:20:31 | Fetching logs for zookeeper/0e82df1a-35e7-49cb-8f2d-de99115c69cd (1): Finding and packing log files (00:00:01)

Task 14 Started  Thu May  3 20:20:31 UTC 2018
Task 14 Finished Thu May  3 20:20:32 UTC 2018
Task 14 Duration 00:00:01
Task 14 done

Downloading resource '613c5683-f7c1-46f4-6f20-abddeee33792' to 'C:\Users\rcw5\go\src\github.com\cloudfoundry\bosh-cli\out\zookeeper.zookeeper.1-20180503-212032-3698219.tgz'...

                                                                           0.00%
Succeeded
```

I will shortly submit a PR to fix this.

rcw5:
Fix #419 by closing the temp file before attempting to move it

### DIFF
--- a/cmd/downloader.go
+++ b/cmd/downloader.go
@@ -71,6 +71,11 @@ func (d UIDownloader) Download(blobstoreID, sha1, prefix, dstDirPath string) err
 		}
 	}
 
+	err = tmpFile.Close()
+	if err != nil {
+		return err
+	}
+
 	err = boshfu.NewFileMover(d.fs).Move(tmpFile.Name(), dstFilePath)
 	if err != nil {
 		return bosherr.WrapErrorf(err, "Moving to final destination")
--- a/cmd/downloader_test.go
+++ b/cmd/downloader_test.go
@@ -76,6 +76,22 @@ var _ = Describe("UIDownloader", func() {
 				Expect(director.DownloadResourceUncheckedCallCount()).To(Equal(0))
 				Expect(fs.FileExists(expectedPath)).To(BeFalse())
 			})
+
+			It("returns error if temp file cannot be closed", func() {
+				fakeFile := fakesys.NewFakeFile("/some-tmp-file", fs)
+				fakeFile.CloseErr = errors.New("fake-err")
+				fs.ReturnTempFile = fakeFile
+
+				director.DownloadResourceUncheckedStub = func(_ string, out io.Writer) error {
+					out.Write([]byte("file-contents"))
+					return nil
+				}
+
+				err := act()
+				Expect(err).To(HaveOccurred())
+				Expect(err.Error()).To(ContainSubstring("fake-err"))
+				Expect(fs.FileExists(expectedPath)).To(BeFalse())
+			})
 		}
 
 		Context("when SHA1 is provided", func() {


Fydon:
Thank you @rcw5.

I'm not sure how best to handle, but, like I mentioned on by [own similar pull request][1], what if there is an error before it reaches the Close? I assume that the deferred RemoveAll will fail? Given that ideally the Close should execute before the Move, but also before the RemoveAll, but I don't know golang well enough to suggest the solution. I previously found [embedded functions][2], but maybe just having two Close statements (where you have it and one in defer) and not checking the deferred one for error.

[1]: https://github.com/cloudfoundry/bosh-cli/pull/171#issuecomment-305245025
[2]: https://play.golang.org/p/Xjy1g6W1Qf

rcw5:
Hi @Fydon. Interesting points. I guess a `defer tmpFile.Close()` could be added immediately before the `RemoveAll()` call. Then if an error occurs due to verification or download failure, it will close the handle before cleaning up the file.

Having said that, I'm not sure of the `RemoveAll()` behaviour on Windows. Perhaps it will close the file beforehand? I'd need to test it out.

dlresende:
Closing and re-opening so that we can have a Tracker story.

idoru:
I tested this out and can verify this problem exists, and the PR fixes the issue.

@rcw5 `RemoveAll()` does not close the file handle, so the `defer tmpFile.Close()` will be needed in order to avoid temporary files accumulating. That problem already exists even without this fix.

mrosecrance:
Merging so that the happy path is fixed. We're happy to accept a second PR for the error case as mentioned above as well.

PR #422, opened 2018-05-07
Print instances/vms tables as they come in

knm3000:
instead of buffering until the end

[#154524580](https://www.pivotaltracker.com/story/show/154524580)

Signed-off-by: Yulia Gaponenko <yulia.gaponenko1@de.ibm.com>

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -451,6 +451,10 @@ func (c Cmd) configureUI() {
 
 		c.deps.UI.ShowColumns(headers)
 	}
+
+	if c.BoshOpts.Parallel > 1 {
+		c.deps.UI.EnableLocking()
+	}
 }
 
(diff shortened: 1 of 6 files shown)


cppforlife:
@knm3000 hmm, now that i think about it, there is still a race to "print" against TaskReporter [1] as it prints begin/end task numbers. i think may be a better approach is to implement `type LockingUI struct` that complies with UI interface [2] and uses a mutex internally.

[1] https://github.com/cloudfoundry/bosh-cli/blob/master/ui/task/reporter.go#L12
[2] https://github.com/cloudfoundry/bosh-cli/blob/master/ui/interfaces.go#L7

knm3000:
@cppforlife I've implemented `type LockingUI struct`, please review.

jfmyers9:

PR #446, opened 2018-06-26
Fix followed logs

kayrus:
Resolves #315

PR #448, opened 2018-06-26
Allow INT signals bypass via ssh

### LINKED ISSUE #447: `bosh logs` doesn't cleanup after itself
Since `bosh logs` uses non-interactive mode, it just kills the local ssh client process, but the server side `tail` process remains running. After certain `bosh logs` commands you'll receive an amount of stale `sudo tail -F` processes on the remove server.

I tried to find a workaround for this issue, and I cannot develop anything better than pseudo-interactive shell...

kayrus:
Resolves #447

additionally this fix allows to run `bosh instance/0 htop`. Maybe it makes sense to obsolete non-interactive part of the ssh subclient, or rewrite ssh part using go functions in order to allow Windows users use bosh cli.

### DIFF
--- a/ssh/non_interactive.go
+++ b/ssh/non_interactive.go
@@ -1,6 +1,8 @@
 package ssh
 
 import (
+	"os"
+
 	bosherr "github.com/cloudfoundry/bosh-utils/errors"
 	boshsys "github.com/cloudfoundry/bosh-utils/system"
 
@@ -28,6 +30,10 @@ func (r NonInteractiveRunner) Run(connOpts ConnectionOpts, result boshdir.SSHRes
 		return boshsys.Command{
 			Name: "ssh",
 			Args: append(append(sshArgs.OptsForHost(host), sshArgs.LoginForHost(host)...), rawCmd...),
+
+			Stdin:  os.Stdin,
+
+			KeepAttached: true,
 		}
 	}
 


kayrus:
Any news on this PR?

belinda-liu:
Hi @kayrus,

Just to clarify, are you only seeing this on Windows? We've been having a little trouble reproducing the initial issue (#447) on our end.

Thanks,
@belinda-liu && @mikexuu, CF BOSH Team

belinda-liu:
Hi @kayrus,

We tried building a CLI with your fix and were unable to see it fix the [issue referenced #447](https://github.com/cloudfoundry/bosh-cli/issues/447). Specifically, after running `bosh logs -f`, we still see `tail -f` processes remain on the server. 
Could you please clarify the steps you took to confirm how the issue was fixed? Also, what version of Bosh CLI are you using?

Thanks,
@belinda-liu && @mikexuu, CF BOSH Team

kayrus:
Hi @belinda-liu,

The issue occurs only when you run `bosh logs -f`.

Regarding the output, I'm not quite sure whether I understand you. How exactly did you test the fix? The `tail: cannot open ‘/var/vcap/sys/log/**/*.log’ for reading: No such file or directory` message was fixed in my previous PR: #446. Did you rebase/merge this PR in order to test the output?

I just rebased this PR on master and tested the fix. It works fine, there are no stale `tail -F` processes left when I press Ctrl+C. However when I use bosh cli without the fix, `tail -F` processes remain on the server when I press Ctrl+C.

belinda-liu:
Hi @kayrus,

Sorry for the confusion! We tested it the same way you described, and we can see that the fix indeed cleans up the stale  `tail -F` processes. We did also test this after we merged in your other PR, which I realize was unclear because we had copied over output from before we merged.

What we meant by changing the output format was that after applying your fix, we still expect each line to be prefixed by the VM name and either `stderr  |` or `stdout  |`. In the example above, this is `zookeeper/31cc1bd1-6b10-4170-9914-0b4b33ce5c52: stderr |` . However, your changes seem to strip this off of the output. Does that help?

kayrus:
@belinda-liu, thanks for the clarification. I'll take a look into this issue.

kayrus:
@belinda-liu, I've updated the PR. Now it should work fine, please test.

belinda-liu:
Hi @kayrus, we've verified that the fix work and will also close #447. Thanks!

PR #455, opened 2018-07-18
Add dav provider (and Artifactory)

titilambert:
Hello, this patch add the webdav support as blob provider.
It has also a special support for artifactory (adding the filepath as a property)
Here an example of `config/final.yml`:
```
---
final_name: myrelease
blobstore:
  provider: dav
  options:
    endpoint: https://my.artifactory.com/artifactory/myrepo/myfolder
    artifactory: true
```
I think I should split the file dav_blobstore.go and put the client part in the **bosh-davcli** project.
But before to do that, I just want to be sure you will accept this new feature.
Could you confirm that ?
If yes how do you want I split this file ?
Of course, I will add some tests...

Thanks !

### DIFF
--- a/releasedir/dav_blobstore.go
+++ b/releasedir/dav_blobstore.go
@@ -1,0 +1,265 @@
+package releasedir
+
+import (
+	gobytes "bytes"
+	//	"crypto/sha1"
+	"encoding/json"
+	"fmt"
+	"io"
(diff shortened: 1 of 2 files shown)


jrussett:
Closing and reopening this PR so that gitbot makes a story for it.

PR #459, opened 2018-07-31
Add --copy flag

muralisc:
https://github.com/cloudfoundry/bosh/issues/1900


voelzmo:
Switched base branch to `develop` as that is where CLI development is done

dpb587-pivotal:
Any ideas on if we can make the flag slightly more explicit on its behavior? To me, `--copy` has ambiguity with "copy my data" or "copy a new disk from this disk". Or maybe that confusion is just me? Other ideas for brainstorming...

 * copy-properties - excludes "cloud" so kind of includes size, too
 * copy-existing-properties - too verbose?
 * copy-metadata - we don't really use the "metadata" term much
 * inherit-properties/metadata
 * disk-properties=copy

More generally, are there other options aside from copy existing properties that we might consider in the future? e.g. `--disk-properties=copy|none|derive`?

muralisc:
Agree. `--copy` alone is ambiguous.
Like the idea of `--copy-properties` and `--disk-properties=<value>`.
We can start with `--disk-properties=copy|default` and add more options down the line.
Preferring `default` over `none` as the db do not let the field to be null.

Issue #414, opened 2018-04-13
`bosh instances --ps` counts all lines as additional instances


voelzmo:
`bosh instances --ps` on my director seems to indicate that it consists of 16 instances, while it should only be 1:

```
$ bosh instances --ps
Using environment '10.0.1.10' as client 'admin'

Task 89. Done

Deployment 'bosh'

Instance                                   Process             Process State  AZ  IPs
bosh/f64b6adf-75ff-417f-9c14-f2c44b0725c8  -                   failing        z1  10.0.1.12
~                                          blobstore_nginx     running        -   -
~                                          credhub             failing        -   -
~                                          director            running        -   -
~                                          director_nginx      running        -   -
~                                          director_scheduler  running        -   -
~                                          director_sync_dns   running        -   -
~                                          health_monitor      running        -   -
~                                          nats                running        -   -
~                                          postgres            running        -   -
~                                          registry            running        -   -
~                                          uaa                 running        -   -
~                                          worker_1            running        -   -
~                                          worker_2            running        -   -
~                                          worker_3            running        -   -
~                                          worker_4            running        -   -

16 instances
```

belinda-liu:
Fixed in #467.

PR #466, opened 2018-08-22
Don't include process rows in instance count

skibum55:
Don't include process rows in instance count. See #414


dpb587-pivotal:
Thanks - will follow up in #467, but FYI, GitHub now supports changing the base branch of a pull request after making it (e.g. `master` to `develop`); edit button by the title should reveal that option.

PR #467, opened 2018-08-22
Don't include process rows in instance count

skibum55:
Don't include process rows in instance count. See #414

### DIFF
--- a/ui/table/table.go
+++ b/ui/table/table.go
@@ -88,6 +88,11 @@ func (t Table) Print(w io.Writer) error {
 	rowCount := len(t.Rows)
 	for _, section := range t.Sections {
 		rowCount += len(section.Rows)
+		for _, r := range section.Rows {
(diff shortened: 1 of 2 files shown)


ui/table/table.go
```diff
@@ -87,7 +87,13 @@ func (t Table) Print(w io.Writer) error {
 	writer := NewWriter(w, "-", t.BackgroundStr, t.BorderStr)
 	rowCount := len(t.Rows)
 	for _, section := range t.Sections {
+		// we're removing process rows from the instance count
```
belinda-liu: This comment seems a bit too specific to instances and processes while the `Table` struct is a more generic structure. I don't think the comment is really necessary since the `if` statement inside the `for` below is fairly clear.

skibum55:
Alright, I rebased and removed the comment.

belinda-liu:
Thanks! Will also close #414.

PR #490, opened 2018-10-30
Add option to rotate certificates during deploy

aeijdenberg:
Our bosh are integrated with Credhub, and we have a number of deployments that are nearly a year old with bosh/Credhub generated certificates that are about to expire.

This patch adds a flag to bosh-cli which rotates the certificates during a 3-pass deployment.

See the [README](https://github.com/govau/bosh-cli/blob/govau/rotation/README.md) for more details and caveats.

We've been able to use this to rotate our cf-deployment certificates without any downtime for either end-users or developer pushing new applications.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"fmt"
+	"os"
 	"path/filepath"
 
 	"github.com/cppforlife/go-patch/patch"
@@ -12,6 +13,7 @@ import (
 	boshtpl "github.com/cloudfoundry/bosh-cli/director/template"
 	boshrel "github.com/cloudfoundry/bosh-cli/release"
 	boshreldir "github.com/cloudfoundry/bosh-cli/releasedir"
+	"github.com/cloudfoundry/bosh-cli/rotation"
 	boshssh "github.com/cloudfoundry/bosh-cli/ssh"
 	bistemcell "github.com/cloudfoundry/bosh-cli/stemcell"
 	boshui "github.com/cloudfoundry/bosh-cli/ui"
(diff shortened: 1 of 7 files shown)


h4xnoodle:
Hello @aeijdenberg, and thank you very much for your contribution!

Certificate rotation is definitely something we've thought a lot about. 

Our approach we've landed on is to have certification rotation originate from the director. There are a few reasons for this, and please let us know if you can think of more. We are looking for input from any users of Bosh, as it's an important feature as you've identified already.

Some of the thinking for why we wish to move certification rotation into the director itself:
1. The director has a full understanding of the 'world', and can open up our options with respect to recovering from partial rotation and recovery from failure.
2. The director understands links, and any cross-deployment links that may be present. We need to carefully rotate in these scenarios. Our plan is to introduce granular APIs for this purpose. For example, operators will need to perform 'step 1' for deployment A, and then step 1 for deployment B. If we chose to have 'one command to rule them all' for cross-deployment links, then his automation could only come from the director.
3. Credhub access is limited to still just the director.
4.  An API/director approach allows for third-party tools that use Bosh to perform this rotation in a more flexible way.
5. The CLI can leverage these director APIs, and continue to be less intelligent about complex Bosh operations.

Our initial thinking is located here, in bosh-notes: https://github.com/cloudfoundry/bosh-notes/blob/master/proposals/certificate-ca-rotation.md (though we're deprecating bosh-notes in favour of well-defined epics in tracker)
Our public tracker is available, and we have the work planned out in this epic: https://www.pivotaltracker.com/epic/show/4121599 (you can click 'add/view stories' to see the stories we will execute on)

We will close this PR now, in favour of developing a director-driven approach. Please let us know if you see anything that can be improved upon in the future!

PR #492, opened 2018-11-08
Add new inspect-local-release command

davewalter:
Hi,

This PR adds a new `inspect-local-release` command, similar to the existing `inspect-local-stemcell` command.

Please let me know if you have any questions.

Regards,
Dave

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -285,6 +285,14 @@ func (c Cmd) Execute() (cmdErr error) {
 	case *InspectReleaseOpts:
 		return NewInspectReleaseCmd(deps.UI, c.director()).Run(*opts)
 
+	case *InspectLocalReleaseOpts:
+		relProv, _ := c.releaseProviders()
+
+		return NewInspectLocalReleaseCmd(
+			relProv.NewArchiveReader(),
+			deps.UI,
+		).Run(*opts)
+
 	case *VMsOpts:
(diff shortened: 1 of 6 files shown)


davewalter:
One thing I did notice tis that the release.MF metadata file inside a release tarball does not include the job-packageName dependencies, so this information is missing from the output, compared to the same output generated by `create-release`. Interestingly, the inter-package dependency information is included in the metadata file, so it is displayed by this new command. I haven't looked into what it would take to add the job-packageName information to the metadata file, but I assume it wouldn't be too hard.

mfine30:
Heya @davewalter, thanks for the PR! I'm curious what the use case for this command vs the `inspect-release` command is?

davewalter:
Hi @mfine30,

Thanks for taking a look at this. I was working with local release tarballs to research an issue with the files our CI was storing in S3, and wanted an easy way to check the version embedded in the metadata file vs the one included in the filename. Uploading the release to a bosh director seemed like overkill, and when I saw that there was an existing `inspect-local-stemcell` command, it seemed logical to add one for releases.

I started out writing my own output, but then I came across the `ReleaseTables` helper and figured it would be nice to get as much output as possible "for free".

Dave

luan:
@davewalter we went ahead and also implemented the `create-release` change to add `packages` to `jobs` so that this command looks more inline with how `create-release` and `inspect-release` present the information.

Thanks for the PR!

PR #506, opened 2018-12-28
Test for source when uploading a release

jaresty:
When uploading releases, the bosh-cli first tests if a release exists
before attempting to upload that release.  Previously, it would consider
a release to exist on the director regardless of whether source bits
exist for that release.  This could result in skipping release uploads
when they actually should be uploaded in the case where the only
releases that are present are compiled.  This commit changes that in the
following way: when checking whether a release has already been uploaded
to the director and no stemcell is specified, we consider if the given
release has any packages that have both a sha1 and a blobstore_id and
therefore have source bits that can be compiled when needed.

### DIFF
--- a/director/releases.go
+++ b/director/releases.go
@@ -127,6 +127,10 @@ type CompiledPackage struct {
 	SHA1        string `json:"sha1"`
 }
 
+func (p Package) HasSource() bool {
+	return p.BlobstoreID != "" && p.SHA1 != ""
+}
+
 func (d DirectorImpl) Releases() ([]Release, error) {
 	var rels []Release
 
@@ -184,7 +188,7 @@ func (d DirectorImpl) HasRelease(name, version string, stemcell OSVersionSlug) (
 		return false, err
 	}
 
-	if !stemcell.IsProvided() || !found {
+	if !found {
 		return found, nil
 	}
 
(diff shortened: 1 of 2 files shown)


jaresty:
The criteria for `HasSource()` for a given package was taken from the PackageValidator class of the director [here](https://github.com/cloudfoundry/bosh/blob/c0eb7d4b48837681c1ef477209fa3012308897e9/src/bosh-director/lib/bosh/director/deployment_plan/package_validator.rb#L18).

pivotal-mp:
We took a look at this and, while it seems to fix the bug at hand, we think it might regress on another use case:
If we're using compiled releases but not specifying the `stemcell` block in the manifest, this will now always re-upload the compiled release, even when it would be duplicate.

Another option we considered: populating each release with one of the stemcell versions specified at the root of the manifest, so there would always be a stemcell specified. The problem here is that we can't easily know which of several specified stemcells would apply to each release, and some of them might be incompatible (e.g. windows stemcell/unix release).

We think the root cause of this confusion is this: the relationship between releases and stemcells configured in the manifest is n:m, and therefore we can never be confident which stemcell a given release will actually work with.

dpb587-pivotal:
IMO the behavior change in this PR is okay with the caveat that we should more strongly recommend the use of `stemcell` in our compiled release-related documentation and the behavior change ought to be mentioned in release notes (I wouldn't consider it a breaking change).

Additional rationales...
 * There's no reasonable way for manifest, CLI, or director to know, in advance, that a release URL refers to a source or specific stemcell. Assuming that, we have to choose what the default assumption ought to be - source or compiled. Because representing a compiled release tarball *requires* more metadata and source should not, I think it's reasonable to assume a URL without additional metadata is source.
 * Short-term, this change may inconvenience deployment authors. However, if they haven't already run into failed deployments of compiled release URLs being skipped in upload due to missing `stemcell`, they're likely to do so in the future (and add `stemcell` at that point).
 * Long-term, this change of obligating `stemcell` should not have any impact on deployment authors - we should be encouraging automation of these compiled release patches, so there should be no additional manual work required.

PR #511, opened 2019-01-15
migrate github.com/jessevdk/go-flags to github.com/cppforlife/go-flags

jtarchie:
This aliasing of `github.com/jessevdk/go-flags` prevents `bosh-cli` repo from being consumed as `go mod` dependency for us.

This still enforces you use of `github.com/cppforlife/go-flags` as the forked usage.

* tests run locally with `/bin/tests`
* adding this line `replace github.com/cloudfoundry/bosh-cli => github.com/jtarchie/bosh-cli v5.4.0+incompatible` to our `go mod` file and running `go get github.com/jtarchie/bosh-cli@master` prevents the same error from happening.

Creating this PR against `develop` branch. Closed #510


luan:
I'm more interested in the team taking the time to reconsider using the fork. It'd be great to migrate back to the upstream if that's possible. I'm not even sure at this point what is the difference on the fork.

I made this so we can look at that soon: https://www.pivotaltracker.com/story/show/164748476

dpb587-pivotal:
Historical context... had the fork made rather than the changes being patched directly in the `vendor` repo like they previously were. The fork is being relied on for building shared objects used by the underlying commands (based on the global arguments); e.g. JSON output, director endpoints. Definitely room for improvement.

jtarchie:
@luan, until then, could you accept this PR and then make the decision about the fork?

terminatingcode:
We'd like to bump this PR as our own migration to go mod has been limited by this forked repo.
We have had success placing a `replace` in our go.mod for the forked repo like so:
```
replace github.com/jessevdk/go-flags => github.com/bosh-dep-forks/go-flags v0.0.0-20170707010757-351f5f310b26
```

However, @jtarchie's solution is far cleaner as far as resolving the forked dependency.

We also have another transitive dependency problem related (`github.com/cheggaaa/pb`) which lacks a go.mod or Gopkg.toml in `v2`.

We may open a separate PR to bump this dependency however as these are two separate issues.

jfmyers9:
Hey all,

I've been looking into this for the past day and I'm not quite sure I understand this pull request. I've successfully created a [bosh client](https://github.com/jfmyers9/bosh-client) that uses `go mod` to vendor the bosh-cli director library as a dependency. The only update I had to make to get it to work was running `go get -u github.com/cheggaaa/pb@v3.0.1` to get `go mod` to use the correct version as a dependency.

Is there anyway you could describe your workflow and why this rename is necessary? Are you trying to vendor other aspects to the CLI in your code?

That being said, I do think we should either update the import path or switch off the fork. I just would like to understand how you all are using the bosh-cli and how this fixes your problem before merging it.

Best,

@jfmyers9

BooleanCat:
Hey there, for the platform recovery case, when we get time in the next few days we can go through the exact migration steps again and copy the output here with our comments.

jfmyers9:
Hi @BooleanCat,

Looking into this more this appears to be due to the fact that you [import `github.com/cloudfoundry/bosh-cli/cmd`](https://github.com/cloudfoundry-incubator/bosh-backup-and-restore/blob/master/factory/bosh_deployment_manager.go#L18) to re-use functionality for reading a CA certificate.

Ideally, you should implement this logic yourself as we do not intend for this package to be imported as a library. We explicitly declare that only `github.com/cloudfoundry/bosh-cli/director` and `github.com/cloudfoundry/bosh-cli/uaa` are supposed to be used as libraries in the README. We would prefer that you did not use functions that we did not explicitly declare as a public interface.

If you remove this import, following the above steps successfully vendor the bosh-cli using `go mod`.

That being said, we are still looking into removing the fork for go-flags.

Best,

@jfmyers9

jfmyers9:
In the meantime, if for some reason you really need to vendor the `bosh-cli/cmd` package for your project, you can use the following replace directive:

```
replace github.com/jessevdk/go-flags => github.com/bosh-dep-forks/go-flags v0.0.0-20170707010757-351f5f310b26
```
This along with `go get -u github.com/cheggaaa/pb@master` will successfully vendor the bosh-cli in your project.

We do hope to switch off the fork soon.

jfmyers9:
Hi all,

We are no longer using the fork for go-flags as of 1df8a67d81693e197eded4a9da0613b084cf5515. Going to close this pull request as a result. Let us know if you have any other issues.

Best,

@jfmyers9

jtarchie:
Thank you!

Issue #519, opened 2019-02-26
Compile CPI failed since PATH is hardcoded


ywei2017:
We tried to upgrade our BOSH environment to the latest one, together with vSphere-cpi-release-51. As it happens, the compiler we have is too old. Since it is a shared jumpbox, we can't just upgrade the system wide compiler. So we installed the compiler in a separate directory.

But the compilation still keeps on failing. After quite some digging, I found that the BOSH CLI uses a [hardcoded path](https://github.com/cloudfoundry/bosh-cli/blob/a18c72307483843b5e190c383ff5e6835b3a8bf2/installation/pkg/compiler.go#L103). Can this be changed so it uses the PATH value from the environment?

Thanks
Yansheng

ywei2017:
See [PR 520](https://github.com/cloudfoundry/bosh-cli/pull/520) for the proposed fix. I already tested locally and it works fine.

jaresty:
Closing this issue since we merged the PR

PR #520, opened 2019-02-27
Update compiler.so so it honors PATH and LD_LIBRARY_PATH env variables

ywei2017:
This is to address [issue 519](https://github.com/cloudfoundry/bosh-cli/issues/519)


ywei2017:
What check is pending, and how long will a status update be available? Just want to see any revision I need for the P/R.

Thanks
Yansheng

jaresty:
This change looks like a good one. We're going to merge it and close the open issue. Thank you!

jaresty && @luan

dpb587-pivotal:
Theoretically, I think this could cause weird edge cases which affect the compiled packages / CPIs. For example, if any CPI dependencies happen to have implicit checks on available libraries and conditionally enable them during compilation (potentially surfacing unexpected dependency errors which are very difficult to reproduce). Additionally, I doubt we pass `LD_LIBRARY_PATH` when invoking the CPI, so if any newly-compiled versions continue to rely on it for runtime they may start to fail.

Just perhaps some things to keep in mind if weird behaviors start showing up.

PR #526, opened 2019-04-09
Update bosh-utils with ssh keep alive

Polar-Beard:
`bosh-utils` was recently updated to send keep alives over the ssh tunnel established between the local machine and the director vm when using `BOSH_ALL_PROXY`: https://github.com/cloudfoundry/bosh-utils/pull/68

We want to pull this change into `bosh cli` since this is the original place we discovered this bug. Essentially, the ssh tunnel can get stale when there's no traffic going through it, which can result in the jumpbox severing the connection and resulting in EOF errors. Traffic tends to not flow when running `bosh create-env` during stages like compilation and setting up the registry. 

We included the keep alive intervals on `bosh ssh` commands as well, since any socks5 tunnel would could potentially be dropped if there's a long period of inactivity. 

We were unsure of how to test this in your repo, and are *extremely* open to suggestions!

cc @notrepo05 @zachgersh


Polar-Beard:
Changes were already made in develop, so this is not needed

PR #536, opened 2019-07-11
Bump pb dependency to latest

BooleanCat:
Bumping this dependency to a later version assists in migrating go projects that depend on bosh-cli to use go modules.

Specifically we'd like to migrate bosh-backup-restore to go modules and cannot at the moment because we depend on bosh-cli and this version of pb causes go modules to misbehave.

Related to #511

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -11,7 +11,7 @@
     "internal/optional",
     "internal/trace",
     "internal/version",
-    "storage",
+    "storage"
   ]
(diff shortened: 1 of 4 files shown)


BooleanCat:
Anything more you need from us to get this merged / scheduled?

jfmyers9:
This has already been merged onto `develop`. Generally we bump dependencies every weekend. Closing this as a result.

PR #537, opened 2019-07-12
Allow tab completion to be more accessible

christarazi:
Users can enable tab completion for the CLI by running:

```
$ eval "$(bosh completion --export)"
```

This utilizes support from the current flag parsing library, go-flags.
The script source can be viewed from go-flags' documentation page:
https://godoc.org/github.com/jessevdk/go-flags#hdr-Completion

Note that this currently only supports bash as the shell.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,12 @@
 
 - [https://bosh.io/docs/cli-v2-install/](https://bosh.io/docs/cli-v2-install/)
 
+## Adding tab completion to bash
+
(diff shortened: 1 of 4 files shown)


christarazi:
It's worth mentioning that to have this persistent across new terminals, reboots, etc. the user is expected to store the output of:

```
bosh completion --export
```

in their `.bashrc` or equivalent.

Issue #549, opened 2020-01-08
allow array indexing lookup of a value


jtarchie:
# Request

With `vars.yml`,

```yaml
people: [ Bob, Susan ]
```

With `test.yml`,

```yaml
someone: ((people.1))
```

When I run the command `bosh int test.yaml -l vars.yml`,
it should output the following:

```yaml
someone: Susan
```

# Currently
When the above is run, it produces an error,

```
Finding variable 'people.1':
  Expected to find a map at path '/1' but found '[]interface {}'
```

This is because `bosh int` assumes all path lookups will be hash-maps.

# Investigation

This appears to be an issue with [bosh-cli creation](https://github.com/cloudfoundry/bosh-cli/blob/6e7218cff8a8dffde32942a9800dcd0df8b5fb3d/director/template/template.go#L178) of a `gopatch.FindOp`.

jtarchie:
I have a PR coming, but after investigation it appears that `go-path` will need to be updated, too.
When switching to `patch.NewPointerFromString`, is appears that an index value always assumes a array, not an array or a hash with that index key.

jtarchie:
After talking to @cppforlife, no changes were made to `go-patch` as this is a very specific feature to bosh interpolation.
The PR #550 reflects this change.

genevieve:
@jtarchie It appears a new release of the bosh-cli was cut with your PR: https://github.com/cloudfoundry/bosh-cli/releases/tag/v6.2.0

We can call this out as a dependency in the readme.

genevieve:
Think this issue can be closed though!

xtreme-conor-nosal:
merged in 6.2.0

PR #550, opened 2020-01-08
support array index when reference a bosh interpolated value

jtarchie:
This is a PR related to this issue:  https://github.com/cloudfoundry/bosh-cli/issues/549

Talked to @xtreme-conor-nosal about this feature.

### DIFF
--- a/director/template/template.go
+++ b/director/template/template.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"regexp"
 	"sort"
+	"strconv"
 	"strings"
 
 	bosherr "github.com/cloudfoundry/bosh-utils/errors"
@@ -179,7 +180,18 @@ func (l varsLookup) Get(name string) (interface{}, bool, error) {
 
 		val, found, err = processPatchPointer(pointer, val)
 		if err != nil {
-			return val, found, err
+			tokens := []patch.Token{}
+			for _, token := range pointer.Tokens() {
+				switch v := token.(type) {
+				case patch.IndexToken:
+					tokens = append(tokens, patch.KeyToken{Key: strconv.Itoa(v.Index)})
+				default:
+					tokens = append(tokens, token)
+				}
+			}
+			pointer = patch.NewPointer(tokens)
+
+			return processPatchPointer(pointer, val)
 		}
 	}
 
@@ -195,7 +207,7 @@ func processPatchPointer(pointer patch.Pointer, val interface{}) (interface{}, b
 	if err != nil {
 		return val, false, err
 	}
-	return newVal, false, nil
+	return newVal, true, nil
 }
 
 type varsTracker struct {
--- a/director/template/template.go
+++ b/director/template/template.go
@@ -172,23 +172,30 @@ func (l varsLookup) Get(name string) (interface{}, bool, error) {
 	}
 
 	if len(splitName) > 1 {
-		tokens := []patch.Token{patch.RootToken{}}
-
-		for _, token := range splitName[1:] {
-			tokens = append(tokens, patch.KeyToken{Key: token})
-		}
-
-		findOp := patch.FindOp{
-			Path: patch.NewPointer(tokens),
-		}
-
-		val, err = findOp.Apply(val)
+		pointer, err := patch.NewPointerFromString("/" + strings.Join(splitName[1:], "/"))
 		if err != nil {
 			return nil, false, err
+		}
+
+		val, found, err = processPatchPointer(pointer, val)
+		if err != nil {
+			return val, found, err
 		}
 	}
 
 	return val, true, err
+}
+
+func processPatchPointer(pointer patch.Pointer, val interface{}) (interface{}, bool, error) {
+	findOp := patch.FindOp{
+		Path: pointer,
+	}
+
+	newVal, err := findOp.Apply(val)
+	if err != nil {
+		return val, false, err
+	}
+	return newVal, false, nil
 }
 
 type varsTracker struct {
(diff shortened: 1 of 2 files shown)


director/template/template.go
```diff
@@ -172,25 +173,43 @@ func (l varsLookup) Get(name string) (interface{}, bool, error) {
...
-
-		findOp := patch.FindOp{
-			Path: patch.NewPointer(tokens),
+		pointer, err := patch.NewPointerFromString("/" + strings.Join(splitName[1:], "/"))
+		if err != nil {
+			return nil, false, err
 		}
 
-		val, err = findOp.Apply(val)
+		val, found, err = processPatchPointer(pointer, val)
```
xtreme-conor-nosal: Is found used after this assignment? It seems redundant if it's only set when err != nil.

director/template/template.go
```diff
@@ -172,25 +173,43 @@ func (l varsLookup) Get(name string) (interface{}, bool, error) {
...
 		if err != nil {
-			return nil, false, err
+			tokens := []patch.Token{}
+			for _, token := range pointer.Tokens() {
```
xtreme-conor-nosal: Replacing all index with key won't work for arbitrary nesting (e.g. if the path has an array index and a numeric key in it).

Looking at the short-circuit code at L170 I'm wondering if a token-by-token approach might be cleaner. Like your suggested go-patch PR that could allow us to reflect on the value for each token and choose the right token, instead of a blind try and retry.

director/template/template.go
jtarchie: Maybe? Golang refactoring did things for me, I'll double check.

director/template/template.go
jtarchie: It turns out that numeric keys aren't a thing with go-patch.
It only looks up keys that are strings.
So that is not an issue, because it never supported it.

For example,

```bash
$ bosh int <(echo "person: ((names.1))") -l <(echo "names: {0: Bob, 1: Susie}")
Finding variable 'names.1':
  Expected to find a map key '1' for path '/1' (found map keys: '')

Exit code 1
```

This is with keys as numbers. It does not work.

When the keys are strings, just happen to be a number:

```bash
$ bosh int <(echo "person: ((names.1))") -l <(echo "names: {'0': Bob, '1': Susie}")
person: Susie

Succeeded
```

It works.

This is current behaviour of the bosh-cli, so this is maintained with the above code.

mxplusb:
I was actually looking for this [today](https://cloudfoundry.slack.com/archives/C02HPPYQ2/p1578522026016100), so the timing here is great!

Why would the array accessor token be `((key.index))` instead of `((key[index]))`? Most programming paradigms are bracket based instead of token based, so the latter would seem/feel a bit more natural. Was it just easier from an implementation standpoint to leverage the former?

jtarchie:
Because I want to do as less work as possible to introduce this feature. Adding square brackets would be a whole new token parsing, which I didn't feel comfortable adding. The dot notation can is similar to the `/` notation used for `path` in ops files.

director/template/template.go
cjnosal: I see, so the current state is:
- go-patch assumes all numbers are array indexes
- bosh assumes all numbers are string map keys
- the current PR tries the go-patch way, then the bosh way

I'd be interested in generalizing the solution to support mixing:
`bosh int <(echo "person: ((0.0))") -l <(echo "'0':['a']")`

But since this is all net new backwards compatible it can be a follow up PR.

xtreme-conor-nosal:
@mxplusb the syntax is derived from https://tools.ietf.org/html/rfc6901#section-3 (as implemented by https://github.com/cppforlife/go-patch) but using '.' as delimiters instead of '/'.

I believe there will be some quirks with this change:
- if the vars file is an array at the root level indexing won't work
- any '~', '/', or ':' that appears after the first '.' will now be parsed by go-patch instead of treated as a literal

Given the two common variable patterns are "/foo/bar" and "a.b.c" I figure those cases aren't prevalent enough to be a concern.

jtarchie:
I can make it not have those side affects.
I can try an confirm with some users if this is the case, too.

director/template/template.go
jtarchie: That becomes a combinatory issue then.
If you'd like that to work, then the request for `go-patch` to support `IndexToken` and `KeyToken` as union type would be better.

jtarchie:
@cjnosal, did you want to follow up with any changes then?

mxplusb:
@jtarchie 

> I can try and confirm with some users if this is the case, too.

I can confirm that my use case is for indexing an array as a root level object.

xtreme-conor-nosal:
Are these changes time sensitive? There's a few angles I'd like to dig into further to have a more generalized "be closer to go-patch" behaviour:

Since root arrays are requested, that would require removing the special case where len(split(.))==1. I think that early return is why the test with slashes still works with your current change, so after removing it I believe we'll need to do the ~ escaping before wrapping the provided path in /'s for go-patch.

I was also experimenting and it looks like go-patch does support quoted numbers as keytokens (e.g. `/foo/'1'` or `/bar/\"2\"` which would let us be unambiguous about whether a number in the path is an index or a key. Unfortunately something in the `director/template` package mutates the literal quotes as part of interpolation so by the time go-patch is searching the key no longer matches the map.

xtreme-conor-nosal:
I did a little more research, the generalizations I was hoping for would be very difficult:
- yaml unmarshaling handles quoted keys, so `"1": a` in the vars file is still just the map key "1", but json path requires byte-identical values, so `/\"1\"` wouldn't match. Working around it would require kludgy string pre-preprocessing (quoted tokens -> unquoted KeyTokens, numbers to IndexTokens) and manually building the token array again
- root level arrays can't be handled with the current architecture. The vars file is unmarshalled as a StaticVariables type (a map). Refactoring to an `interface{}` type or adding some new CLI flag to say "here's an arbitrary yaml file to pull vars from" would be a significant change.

The PR looks good to me

PR #559, opened 2020-05-26
Add --fix-releases flag to the deploy command

beyhan:
Using the new flag `--fix-releases` during deploy will trigger a reupload of all releases form the `releases` section in manifest, which provide an `url`. The reupload will replace the existing releases and is equivalent to the `bosh upload-release --fix` command. This is useful when:
- the upload of a release produced corrupted artifacts in the blobstore
- director blobstore got corrupted
- you need to migrate to a different blobstore

### DIFF
--- a/cmd/cmdfakes/fake_release_uploader.go
+++ b/cmd/cmdfakes/fake_release_uploader.go
@@ -21,6 +21,19 @@ type FakeReleaseUploader struct {
 		result1 []byte
 		result2 error
 	}
+	UploadReleasesWithFixStub        func([]byte) ([]byte, error)
+	uploadReleasesWithFixMutex       sync.RWMutex
+	uploadReleasesWithFixArgsForCall []struct {
+		arg1 []byte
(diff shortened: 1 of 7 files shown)


beyhan:
This functionality could be easily added to the `update-runtime-config` command if it happens to be useful.

aashah:
Looks like the base branch was `master` instead of `develop`. We might want to do something about `ff24766` being present in the branch history now, when it eventually gets merged into master (by CI).

beyhan:
@aashah Thanks for looking into this! Yesterday I had an issue with a blobstore, where I missed this command a lot. Will be nice to get this in. I can add the same support to the `update-runtime-config` so that we have `--fix` support for all possible places, where an upload is possible.

aashah:
Diego and I force pushed to remove the "Merge onto master" commit, as well as clear the conflicts from changing the base branch to develop. Just wanted to let people know why we force pushed :P

PR #561, opened 2020-06-26
Add --fix-releases flag to the update-runtime-config command

beyhan:
Following the pull-request for the deploy command #559 this is adding the new flag --fix-releases to the `update-runtime-config` command. During runtime config upload this flag will trigger a reupload of all releases fr.m the releases section in the runtime-config, which provide an url. The reupload will replace the existing releases and is equivalent to the bosh upload-release --fix command.


beyhan:
@aashah as I already mentioned in #559 I create this pr.

PR #564, opened 2020-07-15
Add start-env and stop-env commands

### LINKED ISSUE #558: [Feature request] Add stop/start-env commands
Deployments created with `create-env` can require maintenance work, during that maintenance the deployment should be in stoped state. E.g. a director deployed with `create-env` and configured with an external DB needs DB upgrades from time to time. During the upgrade the DB has downtime and it's also better to stop the director. That is why `stop-env` and `start-env` commands, which also execute the `drain` step will be very useful for this situation.

Additional context: https://cloudfoundry.slack.com/archives/C02HPPYQ2/p1589402344221100

beyhan:
This pull-request adds `start-env` and `stop-env` commands to the CLI. The commands are very useful when a `create-env` deployment needs maintenance work as DB major update or migration.

**Things to notice**
* the `stop-env` command reuses parts of the `delete-env` command and I notice, that the CLI doesn't execute the `pre-stop` and `post-stop` scripts. I extended the code to execute those scripts, which makes it more compatible with the bosh stop workflow. If you think this is unnecessary we can remove the `pre-stop` and `post-stop` script execution.
* the `stop-env` command doesn't require a `manifest`, but I added it as argument to be consistent with the `start-env` command, which requires a `manifest`.

**Executed tests**
* manual tests
* unit-tests
* integration-tests
* couldn't execute the acceptance tests. 

fixes #558


mrosecrance:
Tested it out and seems to be fine. There are a few typos(`findDeploymend`) in private methods but they'll be easy enough to refactor in the future. When you get the chance, a docs pr here: https://github.com/cloudfoundry/docs-bosh/blob/17267d8e1365fa63a5555d2300a3b73cafdbbf34/content/cli-envs.md would be great with a note about which cli version (v6.3.2+) it will be in.

beyhan:
@mrosecrance Thank you for reviewing and merging! The docs pr is: https://github.com/cloudfoundry/docs-bosh/pull/728

PR #568, opened 2020-09-21
Avoid leaving stray temporary files on windows

idoru:
Currently if any calls between opening and closing `tmpFile` fail, then the deferred `RemoveAll` is triggered. This actually won't work on Windows because the file hasn't been closed yet.

See #421 where this is discussed a bit. I'm hoping this PR can open up a bit of a discussion. Thinking about this was though-provoking for me, at least.

First, you'll note that this doesn't include any tests. Actually the assertions at https://github.com/cloudfoundry/bosh-cli/blob/master/cmd/downloader_test.go#L65-L66 *ought to* have caught this, but because the test employs `FakeFileSystem` from `bosh-utils`, then failure is masked on Windows, because `FakeFileSystem` doesn't emulate Windows filesystem semantics.

Seems like there's a bunch of ways to address this:

0th, do nothing, have no test. Not great.

First, a simple approach could be to avoid `FakeFileSystem` for a single test, which would give test coverage here, but would be a departure from consistency. This would be quick and easy.

Second, make `FakeFileSystem` more robust and emulate Windows filesystem semantics when running on windows (or if feature flagged?). This would be a bit more of an investment, but it could possibly reveal problems in other parts of the codebase (and other codebases) using it.

Now, playing devils advocate for the second case, for a moment. Say if such an investment were to be made, what if it weren't just a question of Windows vs POSIX semantics? Who is to say that a NAS like NFS, CIFS etc might not be the prevailing filesystem on which the software is being run? Now - that's not to say that the code being testing will need to, or indeed has a CI that runs the tests against different OS and/or networked filesystems, but the FakeFileSystem is an abstraction that does make that harder to do, unless it is a master of all.

Which brings me to:

Third, remove file system abstraction from tests. Comes at a higher cost of I/O overhead during testing, possible test pollution.

What do folks think?

### DIFF
--- a/cmd/downloader.go
+++ b/cmd/downloader.go
@@ -54,6 +54,7 @@ func (d UIDownloader) Download(blobstoreID, sha1, prefix, dstDirPath string) err
 		return err
 	}
 
+	defer tmpFile.Close()
 	defer d.fs.RemoveAll(tmpFile.Name())
 
 	d.ui.PrintLinef("Downloading resource '%s' to '%s'...", blobstoreID, dstFilePath)


idoru:
Thought I had signed CLA? Re-opening to see if this fixes it...

PR #573, opened 2021-09-16
Fix disk NeedsMigration to correclty detect differences in Cloud Properties

jpalermo:
The parsing of cloud properties from bosh state json and the deployment manifest
yaml end up giving different object types for slices stored in cloud properties.

It's also likely that numbers might be parsed differently as floats vs ints


rkoster:
LGTM

PR #575, opened 2021-11-18
fix: create-env on ubuntu running in windows wsl2 with virtualbox cpi

nouseforaname:
If using wsl2 virtualbox cant be installed on the linux vm itself

It needs to run on the windows host. The create env would fail because:
The Command Structs UseIsolatedEnv field was hardcoded to true and prevented access to binaries outside the hardcoded PATH. That means that the Windows Native VBoxManage binary could not be accessed by the CPI. This introduces an ENV Var to be able to override the value for UseIsolatedEnv on CPI calls.

For the feature to fully work it requires an additional PR to the vbox cpi. The CPI shells out to the VBoxManage Binary to create a network and uses a "name" parameter. On Windows that value would contain " "(spaces), thus the call failed with too many parameters error on the VBoxManage Binary.

These PRs fix these issues and allow running the virtualbox cpi on wsl2

### DIFF
--- a/cloud/cpi_cmd_runner.go
+++ b/cloud/cpi_cmd_runner.go
@@ -4,6 +4,8 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"os"
+	"strconv"
 
 	bosherr "github.com/cloudfoundry/bosh-utils/errors"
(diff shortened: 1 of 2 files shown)


lnguyen:
LGTM

lnguyen (approved):

PR #580, opened 2021-12-02
Permit --var values to contain newlines.

ragaskar:
- yaml.Unmarshall mangles newlines in non-yaml partials. This commit
  checks the return value type and, if string (what appears to be
  returned by yaml.Unmarshal for invalid YAML input), simply passes
  through the original string value.
- Although this is a behavior change, it is not expected that any
  user would be depending on this behavior.

[#180138088]

Signed-off-by: Rajan Agaskar <ragaskar@vmware.com>

### DIFF
--- a/director/template/var_kv.go
+++ b/director/template/var_kv.go
@@ -33,7 +33,11 @@ func (a *VarKV) UnmarshalFlag(data string) error {
 		return bosherr.WrapErrorf(err, "Deserializing variables '%s'", data)
 	}
 
-	*a = VarKV{Name: pieces[0], Value: vars}
+	if _, ok := vars.(string); ok {
+		*a = VarKV{Name: pieces[0], Value: pieces[1]}
+	} else {
+		*a = VarKV{Name: pieces[0], Value: vars}
+	}
 
 	return nil
 }
--- a/director/template/var_kv.go
+++ b/director/template/var_kv.go
@@ -33,6 +33,8 @@ func (a *VarKV) UnmarshalFlag(data string) error {
 		return bosherr.WrapErrorf(err, "Deserializing variables '%s'", data)
 	}
 
+	//yaml.Unmarshal returns a string if the input is not valid yaml.
+	//in that case, we pass through the string itself as the Unmarshal process strips newlines.
 	if _, ok := vars.(string); ok {
 		*a = VarKV{Name: pieces[0], Value: pieces[1]}
 	} else {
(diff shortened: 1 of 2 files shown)


director/template/var_kv.go
```diff
@@ -33,7 +33,11 @@ func (a *VarKV) UnmarshalFlag(data string) error {
 		return bosherr.WrapErrorf(err, "Deserializing variables '%s'", data)
 	}
 
-	*a = VarKV{Name: pieces[0], Value: vars}
+	if _, ok := vars.(string); ok {
```
klakin-pivotal, line 36: I'm pretty bad at Go, but this is weird enough that it should have a comment explaining what it's doing and why it's doing it. I can _guess_, but I definitely wouldn't be confident in my guess without a lot of research.

selzoc: I think the technical term for this is a [type assertion](https://go.dev/tour/methods/15)

director/template/var_kv_test.go
```diff
@@ -56,5 +56,13 @@ var _ = Describe("VarKV", func() {
 				Expect(val3).To(Equal(map[interface{}]interface{}{"key31": "str"}))
 			})
 		})
+
+		Context("When key/value has newlines", func() {
```
danielfor, line 60: Might as well specify in the context that the string is not a yaml

director/template/var_kv.go
ragaskar: sure, the commit message serves as an explanation, but a comment can be added as a convenience.

director/template/var_kv_test.go
ragaskar: 👍

director/template/var_kv.go
klakin-pivotal: _This_ must be a lie, right?

> If i does not hold a T, the statement will trigger a panic.

rkoster (approved):
LTGM

rkoster:
Enough people have looked at this, gonna merge, Thanks @ragaskar

beyhan:
I somehow missed that I was assigned as reviewer to this pr. I have created a small refactoring suggestion in https://github.com/cloudfoundry/bosh-cli/pull/581. @selzoc and @klakin-pivotal  as you have been involved in the discussion here could you provide your feedback for https://github.com/cloudfoundry/bosh-cli/pull/581.

PR #581, opened 2021-12-08
Try to improve code readability of var_kv

beyhan:
This is a refactoring suggestion for https://github.com/cloudfoundry/bosh-cli/pull/580. I had hard time to read the code and this is a small refactoring to improve that.

### DIFF
--- a/director/template/var_kv.go
+++ b/director/template/var_kv.go
@@ -14,31 +14,32 @@ type VarKV struct {
 
 func (a *VarKV) UnmarshalFlag(data string) error {
 	pieces := strings.SplitN(data, "=", 2)
+	name, value := 0, 1
 	if len(pieces) != 2 {
 		return bosherr.Errorf("Expected var '%s' to be in format 'name=value'", data)
 	}
 
-	if len(pieces[0]) == 0 {
+	if len(pieces[name]) == 0 {
 		return bosherr.Errorf("Expected var '%s' to specify non-empty name", data)
 	}
 
-	if len(pieces[1]) == 0 {
+	if len(pieces[value]) == 0 {
 		return bosherr.Errorf("Expected var '%s' to specify non-empty value", data)
 	}
 
 	var vars interface{}
 
-	err := yaml.Unmarshal([]byte(pieces[1]), &vars)
+	err := yaml.Unmarshal([]byte(pieces[value]), &vars)
 	if err != nil {
 		return bosherr.WrapErrorf(err, "Deserializing variables '%s'", data)
 	}
 
 	//yaml.Unmarshal returns a string if the input is not valid yaml.
-	//in that case, we pass through the string itself as the Unmarshal process strips newlines.
 	if _, ok := vars.(string); ok {
-		*a = VarKV{Name: pieces[0], Value: pieces[1]}
+		//in that case, we pass through the initial flag value as the Unmarshal process strips newlines.
+		*a = VarKV{Name: pieces[name], Value: pieces[value]}
 	} else {
-		*a = VarKV{Name: pieces[0], Value: vars}
+		*a = VarKV{Name: pieces[name], Value: vars}
 	}
 
 	return nil
--- a/director/template/var_kv.go
+++ b/director/template/var_kv.go
@@ -14,32 +14,33 @@ type VarKV struct {
 
 func (a *VarKV) UnmarshalFlag(data string) error {
 	pieces := strings.SplitN(data, "=", 2)
-	name, value := 0, 1
+	const nameIndex, valueIndex = 0, 1
 	if len(pieces) != 2 {
 		return bosherr.Errorf("Expected var '%s' to be in format 'name=value'", data)
 	}
 
-	if len(pieces[name]) == 0 {
+	if len(pieces[nameIndex]) == 0 {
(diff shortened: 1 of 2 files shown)


director/template/var_kv.go
```diff
@@ -14,31 +14,32 @@ type VarKV struct {
 
 func (a *VarKV) UnmarshalFlag(data string) error {
 	pieces := strings.SplitN(data, "=", 2)
+	name, value := 0, 1
```
klakin-pivotal, line 17: These should be constants so that it's clear that they're names for previously-magic numbers. [(Rather than something that is reasonable to change by reassigning to later in the code.)]

I would also suggest calling them "name-index" and "value-index" or something similar. Abbreviating "index" as "idx" would probably be okay.

director/template/var_kv.go
```diff
@@ -14,31 +14,32 @@ type VarKV struct {
...
 
 	//yaml.Unmarshal returns a string if the input is not valid yaml.
-	//in that case, we pass through the string itself as the Unmarshal process strips newlines.
 	if _, ok := vars.(string); ok {
-		*a = VarKV{Name: pieces[0], Value: pieces[1]}
+		//in that case, we pass through the initial flag value as the Unmarshal process strips newlines.
+		*a = VarKV{Name: pieces[name], Value: pieces[value]}
 	} else {
-		*a = VarKV{Name: pieces[0], Value: vars}
+		*a = VarKV{Name: pieces[name], Value: vars}
```
klakin-pivotal, line 42: Why not finish the comment trio and add a comment like `//otherwise, return the parsed YAML object` ?

director/template/var_kv.go
```diff
@@ -14,31 +14,32 @@ type VarKV struct {
...
+	err := yaml.Unmarshal([]byte(pieces[value]), &vars)
 	if err != nil {
 		return bosherr.WrapErrorf(err, "Deserializing variables '%s'", data)
 	}
 
 	//yaml.Unmarshal returns a string if the input is not valid yaml.
-	//in that case, we pass through the string itself as the Unmarshal process strips newlines.
 	if _, ok := vars.(string); ok {
-		*a = VarKV{Name: pieces[0], Value: pieces[1]}
+		//in that case, we pass through the initial flag value as the Unmarshal process strips newlines.
```
klakin-pivotal, line 39: I'm kind of partial to using the word "return", rather than the phrase "pass through"... but I'm _really_ bad at go, so if "pass through" is the phrase that one would expect to be used in this context, then keep using it.

klakin-pivotal (changes requested):
Here are a few changes. Otherwise, it looks fine.

selzoc (approved):

director/template/var_kv.go
beyhan: good suggestion

director/template/var_kv.go
beyhan: comments are always smell that the code is not readable enough but introducing functions here to make comments obsolete will be too much :-). I have added the suggestion.

director/template/var_kv.go
beyhan: "return" is shorter and easier to understand I think. I changed the comment.

nouseforaname:
I added a commit to this PR because we found that the new code also changes rendering behaviour in specific situations

director/template/var_kv_test.go
```diff
@@ -23,6 +23,25 @@ var _ = Describe("VarKV", func() {
...
+				err := (&arg).UnmarshalFlag(`name="val"`)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(arg).To(Equal(VarKV{Name: "name", Value: "val"}))
+			})
+			It("reverts to the old (incorrect) bevaviour and removes the wrapping single quotes if the value is a string", func() {
+				err := (&arg).UnmarshalFlag(`name='val'`)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(arg).To(Equal(VarKV{Name: "name", Value: "val"}))
+			})
+			It("Trim only removes wrapping quotes", func() {
```
klakin-pivotal, line 36: For completeness's sake, can we add in a test or three here that ensures that values with unbalanced quote characters are correctly handled? For example
`"'some-data"'` or `'"some-data'"` get transformed into `'some-data"` and `"some-data'`, respectively?

I expect this to work _today_, but it'd be good to ensure that it keeps working in the future.

beyhan: There was a syntax issue with the test which I fixed. I wanted to add the suggested cases here but they don't work at the moment. You get errors like:
```
  Unexpected error:
        <errors.ComplexError>: {
            Err: <*errors.errorString | 0xc00039ac80>{
                s: "Deserializing variables 'name='\"some-data'\"'",
            },
            Cause: <*errors.errorString | 0xc00039ac50>{
                s: "yaml: found unexpected end of stream",
            },
        }
        Deserializing variables 'name='"some-data'"': yaml: found unexpected end of stream
    occurred
```
Is this expected to work?

nouseforaname: I think it should since that could be a Passwort Value:

`"'someString'"` should be untouched  and render to `"'someString'"` 
`"'someString"` should be touched  and render to `'someString`

PR #584, opened 2022-01-04
Ensure quote stripping behavior for --var matches legacy behaviour.

danielfor:
- We should only remove the first layer of matching quotes.
- We should continue to throw errors when a string cannot be yaml
  unmarshalled. (There hasn't been a request to change this behaviour
  and the impact is unknown)

Signed-off-by: Rajan Agaskar <ragaskar@vmware.com>
Signed-off-by: Long Nguyen <nguyenlo@vmware.com>

### DIFF
--- a/director/template/var_kv.go
+++ b/director/template/var_kv.go
@@ -15,7 +15,6 @@ type VarKV struct {
 func trimWrappingQuotes(s string) string {
 	if (s[0] == '"' && s[len(s)-1] == '"') || (s[0] == '\'' && s[len(s)-1] == '\'') {
 		s = s[1 : len(s)-1]
-		s = trimWrappingQuotes(s)
 	}
 	return s
 }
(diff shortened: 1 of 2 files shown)


klakin-pivotal:
So, if it is true that the old behavior was "Pass what the user gave us directly into `YAML.unmarshal`, and then operate on that datastructure", then what I _really_ want to see here are tests that verify that when a user passes on a variety of strings to this new code, that it behaves just like `YAML.unmarshal` does, except in the one case that we're were trying to fix... namely the one where they're passing something that's _not_ YAML and has embedded newlines in it.

In other words, it would be _very_ good to have a test or tests that feed a bunch of test strings to `YAML.unmarshal`, and ensure that (with the exception of the one case we want to fix), the output from the function that we wrote is the same as what would come out of `YAML.unmarshal`.

Does that seem reasonable and make sense?

klakin-pivotal:
That is what I'm proposing, yes. One big benefit would be to make the intended behavior of the code _very_ clear. All too often, folks who are in a hurry don't have the time to understand _where_ the pre-baked "expected results" data in a test came from... so if it's easy and cheap to be more explicit by creating the expected results at runtime, then I'm always in favor of doing that.

jpalermo (approved):

PR #587, opened 2022-01-28
Fix/add handling for exported_releases in cli

nouseforaname:
This fixes a bug (and somewhat refactors the code) that
was triggered by:
- e75f6a62532e0cea44d60f375fdf168b0bc9c73c
- 0ebe93a21cee221862dadb7ccbdc3b2ae63ba60e

Previously File Urls would always be reuplaoded. Now it will
skip the upload if a name and version in the director match.

that triggered an issue with releases like:

```yaml
- name: example
  url: file:///path.to.tg
  exported_from:
  - os: some
    version: 1.2.3
```

the code to check whether a compiled release needs to be updated
sourced the Stemcell-version for UploadReleaseOpts from the
`ManifestRelease` struct.

The struct unmarshaled the `release: { stemcell: {}}` attribute
in a manifest.

To fix exported_from file releases parsing got added and the
Stemcell Field is populated with the data. This is happening before
parsing of `stemcell`  itself to keep downstream compability

While the manifest item is a list, the docs explicitly state that
we currently only over use the first item and for sake of simplicity
the code to handle this is not introduced with this commit.

Added tests for the behaviour and exposed another interface for more
precise testing.

### DIFF
--- a/cmd/release_manager.go
+++ b/cmd/release_manager.go
@@ -114,7 +114,10 @@ func (m ReleaseManager) createAndUploadRelease(rel boshdir.ManifestRelease) (pat
 		SHA1: rel.SHA1,
 		Fix:  m.uploadWithFix,
 	}
-
+	if len(rel.ExportedFrom) > 0 {
+		// https://bosh.io/docs/locking-compiled-releases/#why-an-array It is an array but we only use the first item.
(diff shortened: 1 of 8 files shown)


lnguyen:
LGTM

PR #590, opened 2022-02-03
Remove registry related code from the cli

nouseforaname:
This PR is the counterpart to removing the Registry from bosh in:
https://github.com/cloudfoundry/bosh/commit/2ba78ed534b764fcf81df4959f2e9a3d59a283bf

Remove Registry Code
Regenerate Mocks

fix minor issues in bin/mockgen (only runs with bsd sed) and remove registry line from it


beyhan (commented):
It looks good. I think that I found one left over where we could rename the registry word or just remove it. It is: https://github.com/cloudfoundry/bosh-cli/blob/0e51615df969861d3e2f39cc73b60ddfb268daf1/integration/create_env_test.go#L97 Otherwise the change looks good to me and sorry for the delayed review.

rkoster:
@nouseforaname could you take a look at the changes requested by @beyhan?

nouseforaname:
> It looks good. I think that I found one left over where we could rename the registry word or just remove it. It is:
> 
> https://github.com/cloudfoundry/bosh-cli/blob/0e51615df969861d3e2f39cc73b60ddfb268daf1/integration/create_env_test.go#L97
> Otherwise the change looks good to me and sorry for the delayed review.

Thanks & done.

beyhan (approved):

jpalermo (approved):
👍

Issue #542, opened 2019-10-04
update-cloud-config does not offer the --named flag


friegger:
https://bosh.io/docs/cli-v2/#update-cloud-config does not support the `--name` flag. With this it cannot do everything that `update-config` can do. As mentioned in https://github.com/cloudfoundry/bosh/issues/2209#issuecomment-538312867 this should actually be the case.

mfine30:
Hey @friegger I agree we should probably add that. It does feel like a good idea so that `update-runtime-config` and `update-cloud-config` match. That being said, I don't think `update-cloud-config` does anything special, so in theory `bosh update-config --name NAME --type cloud` would achieve the exact same result. 

I still see value in this change to help reduce surprises to operators, but not an urgent must have.

mvach:
Hi I'm currently working on that issue.

The director part: https://github.com/cloudfoundry/bosh/tree/update-cloud-config-with-name-param
The cli part: https://github.com/cloudfoundry/bosh-cli/tree/update-cloud-config-with-name-param

I didn't create PRs so far, since there is some more testing that need to be done

mvach:
I created:

- BOSH PR https://github.com/cloudfoundry/bosh/pull/2390
- BOSH-CLI PR https://github.com/cloudfoundry/bosh-cli/pull/600

PR #600, opened 2022-06-28
Add name parameter to cloud configs commands

mvach:
### What is this change about?

This change add a `name` parameter to the commands `update-cloud-config` and `cloud-config`.
This is required to the fix https://github.com/cloudfoundry/bosh-cli/issues/542

In general this solves a inconsistency between the options of command `update-runtime-config` and `update-cloud-config`
This PR depends on: https://github.com/cloudfoundry/bosh/pull/2390

### Please provide contextual information.

https://github.com/cloudfoundry/bosh-cli/issues/542

### What tests have you run against this PR?

I ran the cli unit tests

### How should this change be described in bosh release notes?

The commands `update-cloud-config` and `cloud-config`  offer a `--name` parameter to upload and retrieve a named cloud config

### Does this PR introduce a breaking change?

no

### Tag your pair, your PM, and/or team!
PO: @friegger

### DIFF
--- a/cmd/cloud_config.go
+++ b/cmd/cloud_config.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	. "github.com/cloudfoundry/bosh-cli/v7/cmd/opts"
 	boshdir "github.com/cloudfoundry/bosh-cli/v7/director"
 	boshui "github.com/cloudfoundry/bosh-cli/v7/ui"
 )
@@ -14,8 +15,8 @@ func NewCloudConfigCmd(ui boshui.UI, director boshdir.Director) CloudConfigCmd {
 	return CloudConfigCmd{ui: ui, director: director}
 }
 
-func (c CloudConfigCmd) Run() error {
-	cloudConfig, err := c.director.LatestCloudConfig()
+func (c CloudConfigCmd) Run(opts CloudConfigOpts) error {
+	cloudConfig, err := c.director.LatestCloudConfig(opts.Name)
 	if err != nil {
 		return err
 	}
--- a/cmd/cloud_config_test.go
+++ b/cmd/cloud_config_test.go
@@ -7,6 +7,7 @@ import (
 	. "github.com/onsi/gomega"
 
 	. "github.com/cloudfoundry/bosh-cli/v7/cmd"
+	. "github.com/cloudfoundry/bosh-cli/v7/cmd/opts"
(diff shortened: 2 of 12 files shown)


jpalermo:
I was doing some validation on this and totally thought it was broken because it was always updating the default config no matter what value I used for `--name`. I finally realized I hadn't redeployed the director with the changes I just cut in v273.0.0.

It's not a huge deal, but we at least want to think if we're concerned that this doesn't work with a bosh director <273.

We can:
- Ignore it, it's probably not a huge issue and it will resolve itself given enough time
- Add a check to the cli to throw an error if the director version is <273 (I think this is possible)
- Have the cli abandon the cloud_config api entirely? I'm pretty sure the cloud config and runtime config endpoints were made initially, then the more generic "configs" api was made later when the others showed they were not flexible enough. It seems like it might be possible to just use the configs endpoint directly when doing `cloud-config` based cli commands, but I haven't looked too closely at it. Happy to help with the work if we decide this is the best option.

friegger:
@jpalermo only relying on the configs API is not so easy unfortunately. The `update-runtime-config` command also uploads the releases containing an URL that might be listed in runtime-configs, similar to how the deploy command does it for manifests.

Regarding ignoring vs having a nice error message, for this specific feature I don't expect huge adoption and don't expect many surprises. We probably should add it to the bosh-cli docs [PR](https://github.com/cloudfoundry/docs-bosh/pull/774), though.
But adding the error message could be a good template for any command change, since bosh-cli command changes usually go in sync with Director API changes.

mvach:
Hi @friegger,
I do agree that this is something without huge adoption.
I mean, I implemented it to improve the consistency of `update-runtime-config` and `update-cloud-config`. So no one is really waiting for it.

If it is oke for you and @jpalermo I would update the docs PR as suggested and let the time resolve the issue on the director side.

jpalermo (approved):
Yeah, we could make update-cloud-config use the configs endpoint, but update-runtime-configs would be tricky. At some point we should try to make all of these behave a bit better, but this PR doesn't need to solve these problems and does make the world more consistent.

mvach:
Hi,
I updated the docs as discussed: https://github.com/cloudfoundry/docs-bosh/pull/774/files

friegger:
Looks good then, thanks!

friegger (approved):

PR #601, opened 2022-07-20
allow setting the client field

crhntr:
We would like to customize things in net/http transport layer for an internal CLI tool.


crhntr:
Although it would be nice, I don't think we need this.

PR #603, opened 2022-10-31
add the ability to set a prefix for vendored packages

ramonskie:
this is a temporary solution to solve https://github.com/cloudfoundry/bosh/issues/2400

as we can now add the following to aws cpi for example
`bosh vendor-package ruby-3.1 ~/workspace/bosh/bosh-packages/ruby-release --prefix aws`
this would mean that the vendored package is now available as `aws-ruby-3.1`
so it will not cause any collision with older or newer ruby-3.1 packed for example used by the bosh-director

### DIFF
--- a/releasedir/fs_release_dir.go
+++ b/releasedir/fs_release_dir.go
@@ -236,13 +236,14 @@ func (d FSReleaseDir) BuildRelease(name string, version semver.Version, force bo
 	return release, nil
 }
 
-func (d FSReleaseDir) VendorPackage(pkg *boshpkg.Package) error {
+func (d FSReleaseDir) VendorPackage(pkg *boshpkg.Package, prefix string) error {
 	allInterestingPkgs := map[*boshpkg.Package]struct{}{}
-
 	d.collectDependentPackages(pkg, allInterestingPkgs)
 
 	for pkg2 := range allInterestingPkgs {
+		pkg2.Prefix(prefix)
 		err := pkg2.Finalize(d.finalIndicies.Packages)
+
 		if err != nil {
 			return bosherr.WrapErrorf(err, "Finalizing vendored package")
 		}
--- a/cmd/opts/opts.go
+++ b/cmd/opts/opts.go
@@ -1019,6 +1019,7 @@ type VendorPackageOpts struct {
 	Args VendorPackageArgs `positional-args:"true" required:"true"`
 
 	Directory DirOrCWDArg `long:"dir" description:"Release directory path if not current working directory" default:"."`
+	Prefix    string      `long:"prefix" description:"Prefix to add to the package name" default:""`
 
 	cmd
 }
--- a/cmd/vendor_package.go
+++ b/cmd/vendor_package.go
@@ -33,7 +33,7 @@ func (c VendorPackageCmd) Run(opts VendorPackageOpts) error {
 
 	for _, pkg := range srcRelease.Packages() {
 		if pkg.Name() == opts.Args.PackageName {
-			return dstReleaseDir.VendorPackage(pkg)
+			return dstReleaseDir.VendorPackage(pkg, opts.Prefix)
 		}
 	}
 
--- a/release/pkg/package.go
+++ b/release/pkg/package.go
@@ -17,6 +17,7 @@ func (a ByName) Less(i, j int) bool { return a[i].Name() < a[j].Name() }
 
 type Package struct {
 	resource Resource
+	prefix   string
 
 	Dependencies    []*Package
 	dependencyNames []string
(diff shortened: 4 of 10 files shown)


mvach:
This looks like a nice interim solution for the underlying issue.

rkoster (approved):

jpalermo:
Do the jobs or other packages using vendored package then need to specify `BOSH_PACKAGES_DIR` when sourcing things like [compile.env](https://github.com/bosh-packages/ruby-release/blob/master/src/compile-3.1.env#L4)?

A lot of the core packages support `BOSH_PACKAGES_DIR`, because that's what the cli uses when compiling and using packages locally for a `create-env`, but I don't know if it's ever really been documented as an official "pattern" users should be following.

ramonskie:
that patern doesnt change.
what does change is the package naming now in all those  scripts
for example if i set a prefix `aws` for the ruby-3.1
then they should rename all package locations etc to `aws-ruby-3.1`

beyhan:
@ramonskie you mean to consume this feature a change in this [source command](https://github.com/cloudfoundry/bosh-aws-cpi-release/blob/c7009b2461ab1f9b64adcbe41a3bd048e5e0f108/packages/bosh_aws_cpi/packaging#L7) in the [bosh_aws_cpi](https://github.com/cloudfoundry/bosh-aws-cpi-release/tree/master/packages/bosh_aws_cpi) package of [bosh-aws-cpi-release](https://github.com/cloudfoundry/bosh-aws-cpi-release) will be required or if not could you provide an example?

ramonskie:
here an example for bosh-aws-cpi-release
- run `bosh vendor-package ruby-3.1 ~/workspace/bosh/bosh-packages/ruby-release --prefix aws`
- change 
   https://github.com/cloudfoundry/bosh-aws-cpi-release/blob/master/jobs/aws_cpi/spec#L10 to aws-ruby-3.1
   https://github.com/cloudfoundry/bosh-aws-cpi-release/blob/master/packages/bosh_aws_cpi/spec#L4
   https://github.com/cloudfoundry/bosh-aws-cpi-release/blob/f3f226466800689252f6cb191f4c2a477f908d5f/jobs/aws_cpi/templates/cpi.erb#L6
   https://github.com/cloudfoundry/bosh-aws-cpi-release/blob/25868caf383634f214d239cf889cac82e49506fe/packages/bosh_aws_cpi/packaging#L7
   to `aws-ruby-3.1`

jpalermo:
https://github.com/cloudfoundry/bosh-aws-cpi-release/blob/f3f226466800689252f6cb191f4c2a477f908d5f/jobs/aws_cpi/templates/cpi.erb#L5 might also need to change to an `export`, but my bash knowledge is very non-expert...

jpalermo (approved):
👍

ramonskie:
> https://github.com/cloudfoundry/bosh-aws-cpi-release/blob/f3f226466800689252f6cb191f4c2a477f908d5f/jobs/aws_cpi/templates/cpi.erb#L5 might also need to change to an `export`, but my bash knowledge is very non-expert...

so only line 4 would need a change 
from 
`source ${BOSH_PACKAGES_DIR}/ruby-3.1/bosh/runtime.env`
to
`source ${BOSH_PACKAGES_DIR}/aws-ruby-3.1/bosh/runtime.env`

releasedir/fs_release_dir.go
```diff
@@ -263,8 +263,14 @@ func (d FSReleaseDir) collectDependentPackages(pkg *boshpkg.Package, allInterest
...
-func (d FSReleaseDir) writeVendoredPackage(pkg *boshpkg.Package) error {
-	name := pkg.Name()
+func (d FSReleaseDir) writeVendoredPackage(pkg *boshpkg.Package, prefix string) error {
+	var name string
+	if prefix != "" {
+		name = fmt.Sprintf("%s-%s", prefix, pkg.Name())
+	} else {
+		name = pkg.Name()
+	}
+	// name := pkg.Name()
```
aramprice, line 273: Should this commented code be removed?

aramprice (approved):
LGTM as long as the commented-out code is removed.

releasedir/fs_release_dir.go
ramonskie: nice catch :) will remove it

aramprice: Would you be up for squashing that commit? Would make for cleaner git history.

ramonskie:
put this back to draft. as 1 thing still not handled properly.
when creating a release with the vendored packages
and its finalizing the packages in `.final_builds/packages/`
it sitll uses the original package name and not the prefixed one.
to achieve this it seems we need a bigger refactor of the bosh cli

ramonskie:
and @jpalermo i finally understood what you thought the problem would be.
we need to adjust the ruby-package for example the following

we need to edit the following lines https://github.com/bosh-packages/ruby-release/blob/master/packages/ruby-3.1/packaging#L62-L63
to 
```
envsubst '$BOSH_INSTALL_TARGET' <compile-3.1.env | tee "${BOSH_INSTALL_TARGET}/bosh/compile.env"
envsubst '$BOSH_INSTALL_TARGET' <runtime-3.1.env | tee "${BOSH_INSTALL_TARGET}/bosh/runtime.env"
```
and these lines https://github.com/bosh-packages/ruby-release/blob/master/src/runtime-3.1.env should be
```
export PATH="${BOSH_INSTALL_TARGET}/bin:$PATH"
export GEMRC="${BOSH_INSTALL_TARGET}/bosh/gemrc"
```
and this line https://github.com/bosh-packages/ruby-release/blob/master/src/compile-3.1.env#L4 with
```
source "${BOSH_INSTALL_TARGET}/bosh/runtime.env"
```

jpalermo (approved):
👍

Issue #606, opened 2023-02-02
Updating package via bosh `vendor-package --prefix` does not honor prefix on dependent packages


jrussett:
## Summary

When you **update** a package via `bosh vendor-package --prefix` and the desired package has dependent packages, the dependent packages are not updated with a prefix.

#### Note
On the first invocation of `bosh vendor-package` with `--prefix`, the desired package **and** its dependent packages will all be vendored with a prefix. This error only occurs when re-vendoring the package to update it. 

## Steps to Reproduce
For this specific example, we will use the [`healthchecker` package](https://github.com/cloudfoundry/healthchecker-release/blob/develop/packages/healthchecker/spec#L5), which requires the `golang-1-linux` package to compile the binary. The `healthchecker` package is intended to be vendored into releases like `silk`, `cf-networking`, and `garden-runc`.

When vendoring another bosh package with a PREFIX that depends one of the `golang-*-linux` packages, the `golang-*-linux` will be vendored with prefix (e.g. `PREFIX-golang-*-linux`).

If I run the following
```bash
cd ~/workspace/silk-release
bosh vendor-package healthchecker ~/workspace/healthchecker-release --prefix silk
```

Then I end up with the following packages:
```bash
silk-healthchecker/
silk-golang-1-linux/
```

Where `silk-healthchecker/spec.lock` looks something like:
```yaml
name: silk-healthchecker
fingerprint: 11fec354b909c9d6463e9edd3c92bb7c01423eb0df21c2f7ebe742c5d6f34e28
dependencies:
- silk-golang-1-linux
```

### Running `bosh vendor-package --prefix` again to update the package
Now, when I want to update the `healthchecker` package, I see that the `silk-golang-1-linux` dependency is renamed, removing the prefix entirely!

```diff
diff --git a/packages/silk-golang-1-linux/spec.lock b/packages/silk-golang-1-linux/spec.lock
 name: silk-golang-1-linux
-fingerprint: b55bf89806a951e9b3812e987bd526ef931550ab01360f002b38e00d83f8296b
+fingerprint: db9120088b5e3ad5c6f52742a8f1c21a9b234a3b79cf916d48b22660c7ac422e

diff --git a/packages/silk-healthchecker/spec.lock b/packages/silk-healthchecker/spec.lock
 name: silk-healthchecker
-fingerprint: 11fec354b909c9d6463e9edd3c92bb7c01423eb0df21c2f7ebe742c5d6f34e28
[...]

PR #608, opened 2023-02-14
fix inconsistent vendor package behaviour

nouseforaname:
when we introduced the prefixed vendored package behaviour we overlooked a case that could lead to inconstistent writes of the spec.lock for a prefixed package.

The issue was described in https://github.com/cloudfoundry/bosh-cli/issues/606

The underlying cause was the ordering of `allInterestingPkgs` https://github.com/cloudfoundry/bosh-cli/pull/603/files#diff-420d09437c4bf08b4407d39b5956c6209df3519cde36b121a856085e372d2c92R240-R241

As setup in the added test, we have 3 packages. The top level package `pkg-1` and the two dependencies it pulls in `pkg-2,pkg-3`

The ordering of the map `allInterestingPkgs` is inconsistent and the test will fail for every case where pkg-1 isn't the last item in `allInterestingPkgs`. That is because we end up calling writeVendoredPackage on pkg-1, which in turn will iterate over the dependencies.

At this point we didn't yet call .Prefix and .Finalize on the objects for `pkg-2` and `pkg-3`. Though this is required so they can return their prefixed name via Name().

This is addressed by duplicating the loop. The first pass will ensure all objects in the dependency tree are prefixed and finalized before the spec.lock is written to disk.

### DIFF
--- a/releasedir/fs_release_dir.go
+++ b/releasedir/fs_release_dir.go
@@ -247,8 +247,10 @@ func (d FSReleaseDir) VendorPackage(pkg *boshpkg.Package, prefix string) error {
 		if err != nil {
 			return bosherr.WrapErrorf(err, "Finalizing vendored package")
 		}
+	}
 
-		err = d.writeVendoredPackage(pkg2)
+	for pkg2 := range allInterestingPkgs {
+		err := d.writeVendoredPackage(pkg2)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


ramonskie (approved):
LGTM

rkoster (approved):

PR #609, opened 2023-02-15
ginkgo v2 update

nouseforaname:
As far as I can tell:

units and integration pass when run locally. Didn't run acceptance further than letting it fail with expected messages about config files missing.


aramprice (approved):

selzoc (approved):
:shipit:

PR #610, opened 2023-03-09
Add ability to use logs, ssh, and scp commands with BOSH directors

selzoc:
3 new flags are added to the `logs`, `ssh`, and `scp` commands:
* `--director` targets the director's agent HTTP endpoint
* `--agent-endpoint` corresponds to the agent's HTTP endpoint, reference the [cloud_provider.mbus](https://github.com/cloudfoundry/bosh-deployment/blob/de79d6c790f531b56f32d5f303e9254a8fa1b20b/bosh.yml#L3) property of the director manifest
* `--agent-certificate` corresponds to the certificate provided to the agent's HTTP endpoint, reference the [cloud_prodvider.cert](https://github.com/cloudfoundry/bosh-deployment/blob/de79d6c790f531b56f32d5f303e9254a8fa1b20b/bosh.yml#L2) property of the director manifest

The functionality of all of these commands is the same as without the `--director` flag, with the exception of using a private key/username (which doesn't make much sense in this context).

This flag is `--director` for simplicity, and we imagine that it will mostly be used with BOSH directors, but this feature works with any VM created with the `bosh create-env` command.

This feature depends on `bosh-agent` code that is present on stemcell versions:
* Jammy: `1.93`
* Bionic: `1.181`

### DIFF
--- a/agentclient/mocks/mocks.go
+++ b/agentclient/mocks/mocks.go
@@ -5,37 +5,36 @@
 package mocks
 
 import (
-	reflect "reflect"
-
 	agentclient "github.com/cloudfoundry/bosh-agent/agentclient"
 	applyspec "github.com/cloudfoundry/bosh-agent/agentclient/applyspec"
 	gomock "github.com/golang/mock/gomock"
+	reflect "reflect"
 )
 
-// MockAgentClient is a mock of AgentClient interface.
+// MockAgentClient is a mock of AgentClient interface
 type MockAgentClient struct {
 	ctrl     *gomock.Controller
 	recorder *MockAgentClientMockRecorder
 }
 
-// MockAgentClientMockRecorder is the mock recorder for MockAgentClient.
+// MockAgentClientMockRecorder is the mock recorder for MockAgentClient
 type MockAgentClientMockRecorder struct {
 	mock *MockAgentClient
 }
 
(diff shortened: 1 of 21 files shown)


ragaskar (approved):
I think I follow most of this -- this new functionality depends on a compatible agent, correct? Is there any provision for user understandable errors when running against incompatible agents? 

Maybe something that could be added? Otherwise looks good as far as I can tell.

selzoc:
The error when the agent does not support the command looks like:
```
$ bosh logs --director
Agent responded with error: unknown message bundle_logs

Exit code 1
```

selzoc:
@ragaskar, you did make me think of a situation where an error message could be improved, which is in 6017e1b4320e0011ba6c99d4391c219af738729d.

selzoc:
As an aside, I did find that even with older versions of the agent, you can use `bosh ssh --director` and `bosh scp --director`, since those actions have existed in the agent for a long time.  Only the `bosh logs --director` requires the updated stemcell to function properly.

lnguyen (approved):
LGTM

lnguyen:
@selzoc can you fix conflicts and we can merge after

selzoc:
@lnguyen fixed! :shipit:

Issue #611, opened 2023-03-10
Creating releases with blobs are broken on Windows


selzoc:
When running `bosh create-release` on Windows, the following error is observed:
```
PS C:\Users\cselzo\workspace\windowsfs-online-release> bosh create-release
...
Building a release from directory 'C:\Users\cselzo\workspace\windowsfs-online-release':
  - Moving blob '219376f3-e6b3-4859-4482-beadb66162b9' into cache:
      rename C:\Users\cselzo\.bosh\tmp\bosh-s3-blob3074349478 C:\Users\cselzo\.bosh\cache\sha256:324a029b330444ebae81688f83fce24dbf9253eef6af49d252e5297164f5cd14: The filename, directory name, or volume label syntax is incorrect.
...
```

This is because the blob checksum contains a `:` character (which is a [reserved character](https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file#naming-conventions) in Windows) and the cli [uses the checksum directly](https://github.com/cloudfoundry/bosh-cli/blob/6ddd020005ca246161d200ca97121d0ef4b0e9f2/releasedir/index/fs_index_blobs.go#L139) for the cache filename.

I believe #545 to have the same root cause.

selzoc:
I proposed a fix above, would love to get other folks to take a look to validate my approach.

ramonskie:
to my knowledge sha never uses special characters. so how did a `:` got there in the first place?

selzoc:
@ramonskie https://github.com/cloudfoundry/bosh-cli/commit/5bc33001503ec6f670e73456d531fb8205e715e2 is where this was introduced (when switching from `SHA1` to `SHA256`)

PR #612, opened 2023-03-10
Don't use reserved characters in Windows filenames

selzoc:
Fixes [611](https://github.com/cloudfoundry/bosh-cli/issues/611)


lnguyen (approved):
LGTM

PR #617, opened 2023-03-30
release linux arm64 version

gflex:
add release for linux/arm64
what's missing:
 - homebrew does not support linux on arm64 processors. Hence formula is not updated
 - dockerile for arm64. I'm afraid I'm missing how ci/tasks/create-dockerfile.yml is used for build and if we can provide the target platform as argument.


beyhan:
@rkoster what do you think about this?

rkoster (approved):

cunnie:
Thanks @beyhan — this will help with arm64 Docker images which require the BOSH CLI

PR #620, opened 2023-05-25
Update description for bosh cloud-check command

jpalermo:
Give an example and also note that the flag can be used multiple times.

The docs on bosh.io are more clear, but we wanted to provide a bit of clarity here, too.


cunnie (approved):
Long overdue, thanks!

beyhan (approved):

PR #621, opened 2023-06-10
Better message for failed create-release

cunnie:
When running `bosh create-release` in the wrong directory, the error is cryptic: "Expected non-empty 'name' in config..."

This commit enhances the error message to include the most common cause, being in the incorrect directory, by wrapping the error with "Check that you're in the top-level of the release directory".

Note that although Golang 1.13 introduced error-wrapping functionality, we abide by the pre-1.13 convention, i.e. `bosh.WrapErrorf()`.

Typical error message `cd /tmp; bosh create-release`:
```
Check that you're in the top-level of the release directory:
  Expected non-empty 'name' in config '/tmp/config/final.yml'
```


selzoc (approved):
LGTM 👍

rkoster (approved):

PR #625, opened 2023-08-10
Add `--force-latest-variables` flag to `deploy` command

selzoc:
This will override the behavior of any variables with an `on-stemcell-change` update strategy.  When supplied, the `--force-latest-variables` flag will cause the bosh director to retrieve the latest values of all variables from the config server.

This requires a director with https://github.com/cloudfoundry/bosh/pull/2460 to be effective


selzoc:
Docs PR: https://github.com/cloudfoundry/docs-bosh/pull/800

mvach (approved):
lgtm

PR #627, opened 2023-09-26
feat: Implement pcap cmd

b1tamara:
This PR (still WIP) is an attempt to implement the pcap-lite option _pcap_ action for bosh according to the RFC https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0019-pcap-bosh.md. 

The SSH connection will be used to start the tcpdump tool on the remote VM(s). The data will be transited via the SSH channel back to caller, demultiplexed and written to the file. 

An example for the call: 
`bosh -d haproxy pcap -s=96 -f="host X.X.X.X" -o=test.pcap ha_proxy_z1/d4389c32-9fc2-46dd-be5a-31e6a4a46f31`


b1tamara:
@rkoster

jpalermo:
Hey @b1tamara is this in a state where it could be moved from draft? If so it will get picked up and we can assign some reviewers to take a look at it.

peanball:
Hi @jpalermo, our initial request was for a first and more high level look at this "conceptually" and whether this is abstracted enough already for the BOSH Team or not.

@b1tamara is out for the next couple of days and will be back on Wednesday to continue.

There are still some linter issues that should be fixed before we can consider really reviewing the code.

b1tamara:
Regarding linter check errors: The errors seem to be related to "Unchanged files with check annotations", in other words in the files not related to this PR.

beyhan:
@b1tamara this is it now ready from your side or do you still work on it?

b1tamara:
@beyhan It is ready for review.

domdom82:
regarding the linter issues: I've run golangci-lint version 1.54.2 locally and fixed all issues, however I don't know why the PR validation keeps failing. It is also failing on completely different errors. Seems like there is something wrong with the setup of the validation, can't fix it from here.

cunnie:
I'm not sure this addition to the CLI is necessary.

I think what you're looking to do—capture network traffic for subsequent analysis—can be accomplished with the existing BOSH ssh feature-set. For example, to capture 15 seconds worth of traffic on all the "jammy" BOSH instance groups:

```
bosh -d $DEPLOYMENT -n ssh jammy -c "sudo tcpdump -w /var/vcap/data/out.pcap & sleep 15 ; kill %1"
```

The `.pcap` files generated by `tcpdump` can be retrieved using `bosh scp`:

```
for INSTANCE in $(bosh -d $DEPLOYMENT is --json | jq -r '.Tables[].Rows[] | select(.instance | startswith("jammy")) | .instance'); do
    bosh -d $DEPLOYMENT scp $INSTANCE:/var/vcap/data/out.pcap out-${INSTANCE/\//_}.pcap
done
```

These files can then be [merged](https://www.wireshark.org/docs/wsug_html_chunked/ChIOMergeSection.html) and analyzed with WireShark.

We realize the `bosh scp` command is somewhat cumbersome; however, we believe that an enhancement to `bosh scp` might be better than adding a bespoke `pcap` command to the CLI, since it would facilitate other cases where a user wishes to extract specific files from multiple machines in a deployment.

maxmoehl:
Thanks for raising your concerns @cunnie!

This PR is the result of [RFC0019](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0019-pcap-bosh.md), you can view the full history of options considered in the [pull request](https://github.com/cloudfoundry/community/pull/640). I'll try to summarise the main arguments:

> I think what you're looking to do—capture network traffic for subsequent analysis—can be accomplished with the existing BOSH ssh feature-set. For example, to capture 15 seconds worth of traffic on all the "jammy" BOSH instance groups:
> 
> ```
> bosh -d $DEPLOYMENT -n ssh jammy -c "sudo tcpdump -w /var/vcap/data/out.pcap & sleep 15 ; kill %1"
> ```
> 
> The `.pcap` files generated by `tcpdump` can be retrieved using `bosh scp`:
> 
> ```
> for INSTANCE in $(bosh -d $DEPLOYMENT is --json | jq -r '.Tables[].Rows[] | select(.instance | startswith("jammy")) | .instance'); do
>     bosh -d $DEPLOYMENT scp $INSTANCE:/var/vcap/data/out.pcap out-${INSTANCE/\//_}.pcap
> done
> ```
> 
> These files can then be [merged](https://www.wireshark.org/docs/wsug_html_chunked/ChIOMergeSection.html) and analyzed with WireShark.

We are well aware of this process. The issue is that we have to do it way too often to do it by hand. Building helper scripts proved to be sub-optimal as they only exposed a subset of the functionality of tcpdump / bosh. To mitigate this issue it seemed reasonable to start working on a more resilient, "official" solution (again: see the RFC PR for the full history).

Arguing like this you could question the existence of other commands as well (e.g.: `bosh logs`). Of course there is an alternative route you can go, but we want to establish a resilient, maintained tool that is often needed in our troubleshooting process.

> We realize the `bosh scp` command is somewhat cumbersome; however, we believe that an enhancement to `bosh scp` might be better than adding a bespoke `pcap` command to the CLI, since it would facilitate other cases where a user wishes to extract specific files from multiple machines in a deployment.

This would be a welcome addition in any case!

cunnie:
Hi @maxmoehl , thanks for your detailed response.

> We are well aware of this process. The issue is that we have to do it **way too often** to do it by hand.

Ah—I do packet capture rarely, and usually ssh'ing onto a single VM is adequate. Maybe if I had to do packet capture across multiple VMs frequently I'd feel differently.

rkoster:
The need and scope of the `bosh pcap` command have been discussed in great detail as part of [RFC0019](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0019-pcap-bosh.md), given the RFC was approved and consensus was reached, this PR review should focus on the technical implementation, not the validity of the solution.

pcap/pcap.go
```diff
@@ -0,0 +1,311 @@
...
+	}
+
+	session, err := client.NewSession()
+	if err != nil {
+		client.Close()
+		return nil, fmt.Errorf("ssh: new session: %w", err)
+	}
+
+	tcpdump := addFilterToCmd(tcpdumpCmd, filter, clientSSHAddr.IP.String(), clientSSHAddr.Port)
+	ui.ErrorLinef(tcpdump)
```
ystros, line 166: I like that it prints the `tcpdump` command that it constructed. Makes it more clear to the user what's being run. 👍

cmd/opts/opts.go
```diff
@@ -949,6 +952,20 @@ type RecreateOpts struct {
 	cmd
 }
 
+type PcapOpts struct {
+	Args AllOrInstanceGroupOrInstanceSlugArgs `positional-args:"true"`
+
+	Interface   string        `long:"interface" short:"i" description:"Specifies the network interface to listen on." default:"eth0" required:"false"`
+	Filter      string        `long:"filter" short:"f" description:"Filter to apply when running tcpdump."`
+	SnapLength  uint32        `long:"snaplen" short:"s" description:"Snarf snaplen bytes of data from each packet rather than the default of 65535 bytes." default:"65535"`
+	Output      string        `long:"output" short:"o" description:"File to write pcap to."`
```
ystros, line 961: Currently, if `--output` is not provided to the command, it will connect to VMs and attempt to do the capture, only to fail with `write to output file failed: open : no such file or directory`. It may be better to do some upfront validation that it has been provided and error earlier, similar to what is being done with interface and snap length. Alternatively, you could provide a default output file name if there is one that would make sense.

pcap/pcap.go
```diff
@@ -0,0 +1,311 @@
...
+	if err != nil {
+		return err
+	}
+
+	mergedPackets := mergePackets(packetCs)
+	go func() {
+		for packet := range mergedPackets {
+			err = packetWriter.WritePacket(packet.Metadata().CaptureInfo, packet.Data())
+			if err != nil {
+				ui.ErrorLinef("Writing packet to file failed")
```
ystros, line 122: I'm not familiar enough to know - are errors from the packet writer recoverable? Should we expose the error message here?

ystros (commented):
This is going to be a cool feature! Great job so far.

While testing, I was unable to use the feature against a BOSH environment that we have behind a jumpbox and use `BOSH_ALL_PROXY` to connect to. It failed immediately with a dial timeout error as if it were trying to connect directly from my machine to the internal network IP of the VM. I'm not sure if it was just a peculiarity of that environment, but I didn't see anything in the `pcap` code that would respect it. When I switched to a different environment, it worked as expected. Let me know if I'm missing something.

Other than that, there are couple items below that could help improve the user experience (particularly when things go wrong)

pcap/pcap.go
```diff
@@ -0,0 +1,311 @@
...
+
+	for _, host := range result.Hosts {
+
+		p.ui.BeginLinef("Start capture on %s/%s\n", host.Job, host.IndexOrID)
+
+		packets, err := captureSSH(argv, opts.Filter, username, host, privateKey, opts.StopTimeout, wg, done, p.ui, ctx, cancel)
+		if err != nil {
+			// c.ui.ErrorLinef writes error message to stdout/stderr but does not stop the workflow
+			p.ui.ErrorLinef("Capture cannot be started on the instance: %s/%s due to error. \nContinue on other instances", host.Job, host.IndexOrID)
+
```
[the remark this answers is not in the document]
b1tamara: Good point. I added the error message to the output

cmd/opts/opts.go
b1tamara: I marked `output` as required argument in the command. You are right, it does not make sense to write pcap without output file.

b1tamara:
@ystros 
Thank you very much for your review. Regarding your point to run the pcap feature against a BOSH env behind a jumpbox using gateway flags. Yes, you are right. The current implementation does not support the usage of gateway flags.
The scenario which we want to support on the first place is to start a capture from the jumpbox directly. We are trying to connect to host(s) via ssh connection using the go crypto ssh library.

Supporting gateway flags in order to start the capture using BOSH_ALL_PROXY would make the implementation a bit complicated. The connection to Proxy Server via socks5 protocol should be established and our current ssh connection implementation should be wrapped in. Unfortunately we do not have experience with SOCKS5 protocol / SOCKS5 Proxy and don't have the test environment step up to test such scenarios. Our suggestion would be to start with the feature without supporting gateway flags and would see that as an add on feature.

ystros:
@b1tamara Thanks for the changes!

If the use case is mainly geared towards running on jumpboxes, then this should be fine for now. Support for the gateway flags + BOSH_ALL_PROXY could probably come in a later PR. We should just be sure to clearly document it when `pcap` is added to CLI command docs: https://bosh.io/docs/cli-v2/

ystros (approved):

pcap/pcap.go
```diff
@@ -0,0 +1,311 @@
...
+}
+
+func captureSSH(tcpdumpCmd, filter, user string, host boshdir.Host, privateKeyPem ssh.Signer, stopTimeout time.Duration, wg *sync.WaitGroup, done chan struct{}, ui boshui.UI, ctx context.Context, cancel context.CancelCauseFunc) (<-chan gopacket.Packet, error) {
+	client, err := ssh.Dial("tcp", host.Host+":22", &ssh.ClientConfig{
+		Config:          ssh.Config{},
+		User:            user,
+		Auth:            []ssh.AuthMethod{ssh.PublicKeys(privateKeyPem)},
+		HostKeyCallback: ssh.InsecureIgnoreHostKey(),
+		Timeout:         time.Second,
+	})
```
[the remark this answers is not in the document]
rkoster: Using the `Client` from the `cloudfoundry/bosh-cli/ssh` package should take care of handling the socks proxy: https://github.com/cloudfoundry/bosh-cli/blob/main/ssh/client.go#L122

I would strongly advocate using this ssh Client wrapper, so all ssh connections established by the bosh cli have the same behavior. Feel free to extend the ssh Client wrapper if there are functions of the underlying `golang.org/x/crypto/ssh` Client that are not being exposed.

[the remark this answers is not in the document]
b1tamara: Hello @rkoster ,
We reimplemented the ssh connection creation in order to use boshssh client as you suggested. Our scenario to start the capture directly from jumpbox still works fine. 
However, we need time to investigate how to forward the gateway flags (if set via pcapOptions) to support socks proxy scenario. 
@maxmoehl , as I am OOO next days, feel free to continue on the PR.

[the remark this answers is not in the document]
ystros: Can confirm that the new implementation works with the `BOSH_ALL_PROXY` environments it previously failed in!

ystros (commented):
The updated usage of the BOSH SSH client has resolved the issues with using the feature in environments that rely on `BOSH_ALL_PROXY`. Thanks!

I think the only remaining thing is a couple of linting issues:
```
% golangci-lint run                                                                                                                                                                                 ruby-3.1.4
pcap/pcap.go:163:21: Error return value of `boshSSHClient.Stop` is not checked (errcheck)
		boshSSHClient.Stop()
		                  ^
pcap/pcap.go:169:21: Error return value of `boshSSHClient.Stop` is not checked (errcheck)
		boshSSHClient.Stop()
		                  ^
pcap/pcap.go:179:21: Error return value of `boshSSHClient.Stop` is not checked (errcheck)
		boshSSHClient.Stop()
pcap/pcap.go:185:27: Error return value of `boshSSHClient.Stop` is not checked (errcheck)
		defer boshSSHClient.Stop()
```

jpalermo:
Hey @b1tamara, I think we're ready to merge this in as soon as those linting errors are resolved. Thanks again for this work!

b1tamara:
Hello @jpalermo @ystros. 
Sorry for the response delay, I resolved linter errors and forwarded DisableSOCKS to boshssh.ClientOpts.

rkoster:
Thanks @b1tamara! However there is still one issue left in the linter:
```
  Error: could not import github.com/gopacket/gopacket/pcap (-: # github.com/gopacket/gopacket/pcap
  Error: vendor/github.com/gopacket/gopacket/pcap/pcap_unix.go:35:10: fatal error: pcap.h: No such file or directory
     35 | #include <pcap.h>
        |          ^~~~~~~~
  compilation terminated.) (typecheck)
  ```
  source: [here](https://github.com/cloudfoundry/bosh-cli/actions/runs/6890915681/job/18744901945?pr=627#step:4:73)

**Comment by b1tamara** [comment]
@rkoster I added a new step in the linter workflow to install libpcap-dev for ubuntu runners. It is exciting that now all linter checks are suddenly green.

**Comment by b1tamara** [comment]
The build pipeline might need to be adjusted to install libpcap (the latest version: https://www.tcpdump.org/release/libpcap-1.10.4.tar.gz) otherwise it can lead to compilation failure.

File: cmd/opts/opts.go [hunk 6]
```diff
@@ -949,6 +952,20 @@ type RecreateOpts struct {
 	cmd
 }
 
+type PcapOpts struct {
+	Args AllOrInstanceGroupOrInstanceSlugArgs `positional-args:"true"`
+
+	Interface   string        `long:"interface" short:"i" description:"Specifies the network interface to listen on." default:"eth0" required:"false"`
+	Filter      string        `long:"filter" short:"f" description:"Filter to apply when running tcpdump."`
+	SnapLength  uint32        `long:"snaplen" short:"s" description:"Snarf snaplen bytes of data from each packet rather than the default of 65535 bytes." default:"65535"`
```
domdom82, line 960: `Snarf`?

rkoster (approved):

beyhan:
Thank you for this contribution!

jpalermo:
This was released in 7.5.0

PR #634, opened 2023-10-31
enable global flags for deploy command

fmoehler:


anshrupani (approved):
looks good in general, but having ExcludeDeployments would be good. thanks!

cmd/deploy.go
```diff
@@ -13,24 +14,44 @@ type DeployCmd struct {
...
 	tpl := boshtpl.NewTemplate(opts.Args.Manifest.Bytes)
 
+	configs, _ := c.director.ListConfigs(1, boshdir.ConfigsFilter{Type: "deploy"})
+
+	for _, config := range configs {
+		var conf Conf
+
+		yaml.Unmarshal([]byte(config.Content), &conf)
+		if conf.IncludeDeployments == nil || includeContains(conf.IncludeDeployments, c.deployment.Name()) {
+			c.ui.PrintLinef("Using deployment flags from config of type '%s' (name: '%s')", config.Type, config.Name)
```
mvach, line 49: Can we make that information more prominent in the cli logs (e.g. color it, or add !!! a line before and afterwards) otherwise it can be overseen to easy.

Additionally maybe the better message would be:
NOTE: The deploy command execution will use parameters that are globally configured in the director's %s config.

mvach (changes requested):

cmd/deploy.go
fmoehler: Sure, we can make it more prominent. I tried to have the text similar to what comes up in the deploy command anyway e.g.

Using deployment 'foo'

So I thought it would be nice to stick to the same sentence "structure"

fmoehler:
Documentation PR: https://github.com/cloudfoundry/docs-bosh/pull/811

mvach (approved):

cmd/deploy.go
```diff
@@ -13,24 +14,44 @@ type DeployCmd struct {
...
 }
 
 type ReleaseUploader interface {
 	UploadReleases([]byte) ([]byte, error)
 	UploadReleasesWithFix([]byte) ([]byte, error)
 }
 
+type Conf struct {
+	Flags              []string `yaml:"flags"`
+	IncludeDeployments []string `yaml:"include"`
```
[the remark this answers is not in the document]
fmoehler: Done

danielfor (approved):
LGTM

cmd/deploy.go
```diff
@@ -13,24 +14,58 @@ type DeployCmd struct {
...
+		deploymentIncluded := containsDeployment(conf.IncludeDeployments, c.deployment.Name())
+		deploymentExcluded := containsDeployment(conf.ExcludeDeployments, c.deployment.Name())
+
+		if conf.ExcludeDeployments != nil &&
+			conf.IncludeDeployments != nil {
+			c.ui.PrintLinef("Ignoring deployment flags from config of type '%s' (name: '%s'). Please use only 'include'- OR 'exclude'-property in the config.", config.Type, config.Name)
+		} else {
+			if (conf.IncludeDeployments == nil && conf.ExcludeDeployments == nil) ||
+				deploymentIncluded ||
+				(!deploymentExcluded && conf.ExcludeDeployments != nil) {
```
bgandon, line 59: I first thought of such pseudo-code instead:
```
decision=NO
IF includeList == null OR includeList contains deploymentName THEN decision=YES
IF excludeList != null AND excludeList contains deploymentName THEN decision=NO
```

Then, inspired by the [director’s code for addons](https://github.com/cloudfoundry/bosh/blob/main/src/bosh-director/lib/bosh/director/addon/addon.rb#L39-L42), here is a better version (see other commend below), which is equivalent:
```
decision = applies(includeList, deploymentName) && !applies(excludeList, deploymentName)
```
This is easier to understand, and consistent with what we already have for other include/exclude mechanisms in Bosh.

cmd/deploy.go
```diff
@@ -13,24 +14,58 @@ type DeployCmd struct {
...
+				c.ui.PrintLinef("!!!!!!!")
+				c.ui.PrintLinef("Using deployment flags from config of type '%s' (name: '%s')", config.Type, config.Name)
+				c.ui.PrintLinef("!!!!!!!")
```
bgandon, line 62: Message should be improved to be similar to other Bosh CLI outputs and still raise the relevant attention from human operators

cmd/deploy.go
```diff
@@ -77,6 +112,32 @@ func (c DeployCmd) Run(opts DeployOpts) error {
...
+			opts.Recreate = true
+		case "recreate-persistent-disks":
+			opts.RecreatePersistentDisks = true
+		}
+	}
+
+	return opts
+}
+
+func containsDeployment(slice []string, value string) bool {
```
bgandon, line 132: Thinking of the [implementation done in Director for addons](https://github.com/cloudfoundry/bosh/blob/main/src/bosh-director/lib/bosh/director/addon/addon.rb#L39-L42), I think that `applies()` would be better name for this function, with this pseudo code as implementation:
```
list == null OR list contains deploymentName
```

bgandon (changes requested):

cmd/deploy.go
fmoehler: @bgandon thanks for the review. I agree that would be easier to understand, however only having the check 
`decision = applies(includeList, deploymentName) && !applies(excludeList, deploymentName)`

would not apply to the following cases, that are currently taking care of:
1. No warning/ignoring of the flags in case both include and exclude properties are maintained.
2. If the exclude property is empty, it will always be a NO decision (considering that the containsDeployment() -> applies() method gets rewritten to return true if `list == null OR list contains deploymentName`)

So I am not sure about that...

bgandon: That’s right, my second suggestion is not equivalent to the first one, and i didn’t take into account the ambiguous case where both include and exclude are specified.

The idea with my first suggestion, was to skip the ambiguity and just le the exclude win when both are contradictory concerning a given deployment name.

I think we can go ahead with the code as is.

bgandon (approved):

beyhan:
The lint issues are fixed now and I'm going to merge this as discussed during the FI WG meeting.

jpalermo:
This was released in 7.5.0

PR #635, opened 2023-11-28
Avoid `gopacket/pcap` import to ease builds

jpalermo:
We want to avoid importing `gopacket/pcap`; it forces CGO_ENABLED=1, which makes cross-compilation more challenging, which makes publishing binaries more difficult.

fixes, when running `bin/build`:

```
vendor/github.com/gopacket/gopacket/pcap/pcap.go:30:22: undefined: pcapErrorNotActivated
```

fixes, when running acceptance tests:

```
../vendor/github.com/gopacket/gopacket/pcap/pcap_unix.go:35:10: fatal error: 'pcap.h' file not found
```


cunnie (approved):
LGTM! Fewer lines == better code

PR #638, opened 2023-12-12
When debugging requests, do not dump the body of the request.

jpalermo:
We had a report and stack trace that showed the cli running out of memory when the request body is dumped while uploading a stemcell.

While it would be good to keep this behavior for non-binary request bodies, our tests showed that those don't get printed out anyway. So just changing the true to a false in the hopes that it resolves the OOM issue.


cunnie (approved):
LGTM

PR #641, opened 2024-02-07
`repack-stemcell` stops prefixing archive members with "./"

selzoc:
Prior to this commit, running `bosh repack-stemcell` would result in a repacked stemcell with archive members:
```
./
./dev_tools_file_list.txt
./image
./stemcell.MF
./packages.txt
```

while the original stemcell has archive members:
```
dev_tools_file_list.txt
image
packages.txt
stemcell.MF
```

This commit preserves the original stemcell archive members in the repacked stemcell.  This is useful if you're trying to extract a particular file from the archive by name.

In addition, this commit:
- Vastly simplifies the repack-stemcell command tests (including removing an entire compressor fake separate from the existing one in bosh-utils)
- Removes unneeded UI and FS depedencies from the repack-stemcell command

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -209,7 +209,7 @@ func (c Cmd) Execute() (cmdErr error) {
 		stemcellReader := bistemcell.NewReader(deps.Compressor, deps.FS)
 		stemcellExtractor := bistemcell.NewExtractor(stemcellReader, deps.FS)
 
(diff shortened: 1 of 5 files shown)


ystros (approved):

selzoc:
@jpalermo you're listed as a reviewer here, can you take a look and/or merge?

PR #644, opened 2024-03-20
Remove PowerDNS dependent `--dns` flag and related functionality

aramprice:
PowerDNS is being removed from the BOSH Director, and as a consequence the data returned from the `instances` and `vms` commands no longer have DNS data returned.

This commit removes the `--dns` flag and logic to display the data returnd when it is sent.

See: https://github.com/cloudfoundry/bosh/pull/2501

### DIFF
--- a/cmd/instance_table.go
+++ b/cmd/instance_table.go
@@ -32,9 +32,6 @@ type InstanceTableValues struct {
 
 	Stemcell boshtbl.Value
 
-	// DNS
-	DNS boshtbl.Value
-
 	// Vitals
 	Uptime boshtbl.Value // only for Process
 	Load   boshtbl.Value
@@ -76,9 +73,6 @@ var InstanceTableHeader = InstanceTableValues{
 
 	Stemcell: boshtbl.NewValueString("Stemcell"),
 
-	// DNS
-	DNS: boshtbl.NewValueString("DNS A Records"),
-
 	// Vitals
 	Uptime: boshtbl.NewValueString("Uptime"), // only for Process
 	Load:   boshtbl.NewValueString("Load\n(1m, 5m, 15m)"),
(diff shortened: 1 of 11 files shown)


cmd/instance_table.go
```diff
@@ -155,9 +149,6 @@ func (t InstanceTable) ForVMInfo(i boshdir.VMInfo) InstanceTableValues {
 
 		Stemcell: boshtbl.NewValueString(stemcell),
 
-		// DNS
-		DNS: boshtbl.NewValueStrings(i.DNS),
```
ystros, line 159: `i` here is a VMInfo struct, which still includes the `DNS` field returned from the Director instance info API call. It's still defined on this PR.

https://github.com/cloudfoundry/bosh-cli/blob/425670d1ae381515064add395774bb12fb212c0d/director/vms.go#L26

The only places `vmInfo.DNS` is referenced now appear to be tests.

* Should `VMInfo#DNS` be removed as part of this PR?
* On the flip side, what will happen when someone uses an older version of the CLI that still expects the `dns` field against a newer Director version that doesn't return it? (more relevant to the Director PR, but trying to understand if we're introducing compatibility issues. https://github.com/cloudfoundry/bosh/pull/2501/files#diff-5582c140ad3a74b22a7e4323ac5ff4a237da411c5b531c7da1face14c6c2998aL68)

aramprice: Agreed, removing.

aramprice: With this change the CLI loses the ability to display a response that contains DNS info, however it also loses the ability to request a response containing that info (removal of the `--dns` flag). So I think the CLI is fine.

Updating https://github.com/cloudfoundry/bosh/pull/2501 with the question about director compatibility with a CLI which sends `--dns`

lnguyen (approved):
LGTM

ystros (approved):

PR #647, opened 2024-05-03
Update auto-release pattern

ystros:
Previously, if the patch version of the CLI was bumped due to the `check-for-updated-binary-version` task returning successfully, then the `check-for-patched-cves` task afterward would fail because it would try to look up the new version as a git tag. If that didn't cause a failure and it also returned successfully, the version could have been bumped twice.

Now, the golang-release task scripts create a
release-notes/needs-release file when the condition for the script is met. This allows us to chain the scripts together, build up release notes from each script, then bump a single time if any of the scripts were successful.


ystros:
See also: https://github.com/cloudfoundry/bosh-package-golang-release/commit/be6d9118f0a12a9db822e237237cc9c2e732c611

aramprice (approved):
seems reasonable...

PR #652, opened 2024-05-30
Allow usage of pre-compiled CPI for create-env command

selzoc:
The machinery to allow the use of pre-compiled releases for things _other_ than the CPI has existed for some years.

Note that the CPI must be compiled for the architecture and OS of the machine you are running the create-env command from.  At the time of writing, we're not sure if that architecture and OS must match that of the BOSH director you're creating.

### DIFF
--- a/installation/pkg/compiler.go
+++ b/installation/pkg/compiler.go
@@ -1,17 +1,19 @@
 package pkg
 
 import (
+	"fmt"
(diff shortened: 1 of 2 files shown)


lnguyen (approved):
LGTM

aramprice (approved):

PR #653, opened 2024-05-31
allow white-space in (( var_name )) template

nouseforaname:
currently

`bosh int <( echo "test: (( var_name ))")  --v=var_name=value` will render

```
test: (( var_name ))
```

with the provided patch it will render:

```
test: value
```

I guess this behavior has been the same forever and this is not strictly required. I just happened to lose 20 minutes of my life trying to figure out why my interpolation didn't work and figured I'd at least try to create a commit.


aramprice (approved):
seems like a reasonable change though I'm not super familiar with the templating code...

PR #664, opened 2024-10-02
Pass existing disk CIDs to CPI in create-env command

selzoc:
The create_vm (ref https://bosh.io/docs/cpi-api-v2-method/create-vm/) CPI method takes an optional array of CIDs that feed into the VM placement strategy.  The BOSH director sends these along, but the create-env command did not.  This commit adds that functionality by reading the disks from the state file.


xtreme-nitin-ravindran (approved):
All unit-tests passed. LGTM

PR #669, opened 2024-10-16
Use multi-core optimized pgzip package for tar traversal

rkoster:
This PR is blocked by: https://github.com/cloudfoundry/bosh-utils/pull/98


director/release_archive_with_metadata.go
```diff
@@ -2,7 +2,7 @@ package director
 
 import (
 	"archive/tar"
-	"compress/gzip"
+	gzip "github.com/klauspost/pgzip"
```
aramprice, line 5: same as in `bosh-utils`: this import alias only hides info, would recommend removing it.

also putting non-standard-lib imports here will make the linter 😡

rkoster:
Blocked on bosh-utils ci being broken

jpalermo (approved):

PR #670, opened 2024-10-30
Fix create-env bug where an empty slice of disks from the state file results in null being passed to the CPI ask the disks argument

jpalermo:
This is caused by the way the original string slice is created. If you create it using var []string, it json marshals as null Instead we need to make([]string, 0), which will json marshal as []

### DIFF
--- a/cmd/create_env_test.go
+++ b/cmd/create_env_test.go
@@ -1,6 +1,7 @@
 package cmd_test
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


selzoc (approved):

jpalermo:
Error seen using the vsphere cpi before this change:
```
Deploying:
  Creating instance 'bosh/0':
    Creating VM:
      Creating vm with stemcell cid 'sc-ca101912-4a4c-434a-9e7a-5ce2e426f6a4':
        CPI 'create_vm' method responded with error: CmdError{"type":"Unknown","message":"undefined method `map' for nil","ok_to_retry":false}

Exit code 1
```

Issue #660, opened 2024-08-28
Isolated environment is hard-coded for building bosh-releases


joergdw:
The bosh-cli sometimes compiles releases, e.g. for `bosh create-env`. It therefore [requires](<https://bosh.io/docs/cli-v2-install/#additional-dependencies>) compilation tools that may need certain environment-variables, e.g. to find dynamically linked libraries and header-files.

[installation/pkg/compiler.go:95-108](<https://github.com/cloudfoundry/bosh-cli/blob/db8be01534741118b6787668aa9d985fa7f93a6d/installation/pkg/compiler.go#L95-L108>) prohibits the propagation of needed environment-variables that are intended to solve this. One examle is [`LIBRARY_PATH`](<https://gcc.gnu.org/onlinedocs/gcc-12.4.0/gcc/Environment-Variables.html>) which is like the (already propagated) `LD_LIBRARY_PATH` but relevant for compilation-time instead of run-time. But there are quiet some more.

For me it would be sufficient, to switch off the [`UseIsolatedEnv`](<https://github.com/cloudfoundry/bosh-cli/blob/db8be01534741118b6787668aa9d985fa7f93a6d/installation/pkg/compiler.go#L106>) e.g. via a cli-flag. Then the caller is responsible for providing a correct and clean environment.

rkoster:
bosh releases should be self-contained and should bring the needed header-files / libraries. For example, the ruby release bring libyaml: https://github.com/cloudfoundry/bosh-package-ruby-release/blob/main/packages/ruby-3.3/packaging#L55

joergdw:
In case of using isolated environments, bosh could use a docker-image that provides everything it needs for its release-compilations. This would be then truly-isolated and reproducible at least on all systems that provide a running docker-demon.

jpalermo:
We do provide stemcells as docker images, so those might be a good option for running a create-env from: https://github.com/cloudfoundry/bosh-linux-stemcell-builder/pkgs/container/ubuntu-jammy-stemcell

FloThinksPi:
> We do provide stemcells as docker images, so those might be a good option for running a create-env from: https://github.com/cloudfoundry/bosh-linux-stemcell-builder/pkgs/container/ubuntu-jammy-stemcell

[...]

PR #673, opened 2024-11-19
Remove isolated environment from cpi and compiler cmd

fmoehler:
> This is the same as PR https://github.com/cloudfoundry/bosh-cli/pull/663 . The creator of the PR is currently OOO. Therefore this PR is opened.

The bosh-cli wipes parts of the environment provided by the host-system which are in some cases essential for compilation of bosh-releases. This PR follows a conservative fix-approach by adding an option to opt-out of this default. This can (and maybe should) be extended by a fix of the “isolation” from the host-environment. For more details, see https://github.com/cloudfoundry/bosh-cli/issues/660

Edit: We agreed to remove the isolated environment entirely


aramprice:
I'd prefer to keep this complexity out of the bosh cli since the issue this appears to be addressing could also be side-stepped by running the bosh CLI in a [docker container](https://github.com/cloudfoundry/bosh-cli/issues/660#issuecomment-2377248362).

Perhaps there is some broader issue I'm missing here though?

Regarding the code changes, it seems like there are a number of additions, specifically for nix, beyond a "don't isolate my environment" flag, I'm wondering if those need to be in the `bosh-cli` repo?

Regarding the feature I am concerned that allowing more variation in the local environment will increase the surface area we will have to consider when triaging issues with the CLI.

cloud/cpi_cmd_runner.go
```diff
@@ -100,26 +102,19 @@ func (r *cpiCmdRunner) Run(context CmdContext, method string, apiVersion int, ar
...
-	}
 
 	cmdPath := r.cpi.ExecutablePath()
 	cmd := boshsys.Command{
 		Name: cmdPath,
 		Env: map[string]string{
 			"BOSH_PACKAGES_DIR": r.cpi.PackagesDir,
 			"BOSH_JOBS_DIR":     r.cpi.JobsDir,
-			"PATH":              "/usr/local/bin:/usr/bin:/bin:/sbin",
+			"PATH":              os.Getenv("PATH"),
```
aramprice, line 112: Perhaps a default value of `"/usr/local/bin:/usr/bin:/bin:/sbin"` should remain here and we allow an override with something like `BOSH_CPI_COMPILE_ PATH`.

aramprice:
I think there are two core changes lurking in this PR
1. Allowing the current `ENV` to be passed to the CPI command
  - my inclination is for this logic to live _exclusively_ in the bosh-cli codebase, and not rely on the bosh-utils' `system.execCmdRunner.buildComplexCommand()` handling of `cmd.UseIsolatedEnv` at all.
> It seems like this change alone might be sufficient for the requestors needs.

2. Allowing the the `PATH` value passed to the CPI to be overridden
  - This could be handled by using `BOSH_CPI_COMAND_PATH` or defaulting to the existing value. However this is accomplished the CLI's default behavior should be the same (e.g. `PATH=/usr/local/bin:/usr/bin:/bin:/sbin`).
> This might be a nice additional change _if_ there is a need for a _mostly_ isolated ENV except with modifications to `PATH`.

The other `nix` specific files don't belong in this repo.

Side note: bosh-cli appears to be the only project in the `cloudfoundry` org which sets `system.Command.UseIsolatedEnv` perhaps this capability could be removed if we hoist the "isolated or not" logic into the bosh-cli?

jpalermo:
We talked about this at the FIWG meeting.

We think a good direction is to just always enable the isolated environment within the bosh cli. Even removing the current flag that allows you to disable it. And then have the CLI pass in the full set of ENV variables that a CPI might need to run (PATH, LIBRARY_PATH, etc)

This might cause some breakages, but we can just deal with those if they arise.

jpalermo:
@beyhan and @rkoster , do either of you remember what we talked about in the FIWG meeting regarding this? I can't find the video for the 12/19 meeting and my comment above only confuses me.

I think what my comment is saying is: "Always enable isolated environment, but then pass through ENV that includes PATH and things that might be needed"

But wouldn't it be more simple if we instead "Always disabled isolated environment, and then just let it use whatever is already in the ENV"?

beyhan:
> But wouldn't it be more simple if we instead "Always disabled isolated environment, and then just let it use whatever is already in the ENV"?

@jpalermo I can't follow what you mean. If we don't want to relay on the `bosh-utils` option `UseIsolatedEnv` we could just define a new option with `UseHostOSEnv` or `UseOSEnv` (because UseIsolatedEnv is a little bit confusing IMO and doesn't describe what happens really) and based on that option we decide in https://github.com/cloudfoundry/bosh-cli/blob/main/cloud/cpi_cmd_runner.go#L112-L122 whether we initialise the `env` as it is happening currently or use the OS env as done here https://github.com/cloudfoundry/bosh-utils/blob/c837f5a2ec9037ca1459b49dee470b9d6e4014a5/system/exec_cmd_runner.go#L85. Personally, I would prefer a solution which doesn't change the current behaviour.  
 
If allowing an override of the passed `PATH` value to the CPI is still needed we can talk about introducing a `BOSH_CPI_COMAND_PATH` as Aram suggest but it is not clear to me whether it is needed if OS env usage is implemented.

jpalermo:
I'm just wondering if we actually get anything of value from `UseIsolatedEnv`. 

We've had it configured this way [since 2014](https://github.com/cloudfoundry/bosh-cli/commit/2cfd3a4087fb98322c00846b79f6042244de7288), but it's not clear why we're doing this. The original commit doesn't describe any problem that it is solving. It seems like more of a "this sounds like a good idea" sort of thing.

Up until now, it hasn't really caused any major problems, other than the PR in 2022 to add the flag to disabled parts of isolated env (but PATH is still overridden).

But the BOSH_CPI_USE_ISOLATED_ENV and this new PR kind of show that the IsolatedEnv is problematic. And if we can't describe why we need it, does it really make sense to be adding a bunch of flags and conditional behavior to work around it?

Why not just rip isolated env out of the bosh-cli entirely?

fmoehler:
I also agree with @jpalermo , but am just wondering if removing this entirely could be a security risk? Don't really know if that  is something to consider, just came to my mind.
If it is (or could be), then I think the initial solution (also including this [change](https://github.com/cloudfoundry/bosh-cli/pull/673/files#r1881243425)) to "hide" this feature behind a flag as proposed here is fine.

jpalermo:
Update from FIWG meeting today. Plan is to just remove isolated environment from the cli and let it use whatever is in the ENV

cmd/opts/opts.go
```diff
@@ -197,6 +197,7 @@ type CreateEnvOpts struct {
 	StatePath               string `long:"state" value-name:"PATH" description:"State file path"`
 	Recreate                bool   `long:"recreate" description:"Recreate VM in deployment"`
 	RecreatePersistentDisks bool   `long:"recreate-persistent-disks" description:"Recreate persistent disks in the deployment"`
+	AvoidIsolatedEnv        bool   `long:"avoid-isolated-environment" short:"I" description:"Compile and run cpi-commands in a clean environment or not"`
```
jpalermo, line 200: This should be removed too right?

aramprice (approved):
Adding a +1 to @jpalermo's comment about removing `AvoidIsolatedEnv bool`.

Looks good otherwise, thank you for the contribution!

cmd/opts/opts.go
fmoehler: Of course! Thanks for pointing out. Removed!

jpalermo (approved):

PR #675, opened 2024-12-09
add integration tests for finalize release using bosh-cli

nader-ziada:
These test cases used to be part of the Bosh Director repo when the CLI was included there. But now that CLI v2 is separate and finalize doesn't use director, it's better to move them here.


nader-ziada:
creating as draft as I'm still adding test cases for the `create-release` test

jpalermo:
@nader-ziada could you link to the commits where this was removed from the director?

rajathagasthya (approved):

aramprice (approved):

aramprice:
Here's the existing test we hope to delete after this is merged:
https://github.com/cloudfoundry/bosh/blob/main/src/spec/integration/release/create_release_spec.rb

aramprice:
@nader-ziada / @rajathagasthya can these two commits be squashed, or should is there context in having both of them?

selzoc:
> Here's the existing tests we hope to delete after this is merged: https://github.com/cloudfoundry/bosh/blob/main/src/spec/integration/release/create_release_spec.rb https://github.com/cloudfoundry/bosh/blob/main/src/spec/integration/release/finalize_release_spec.rb

These have been removed as of https://github.com/cloudfoundry/bosh/commit/ad4b9c5f2d09893e1c491f1616f96cd60f2fb24b

PR #684, opened 2025-03-03
Add linters

rajathagasthya:
Adds exceptions in some cases.

Linters not enabled need to be fixed at a later time after examining changes.

### DIFF
--- a/acceptance/cmd_runner.go
+++ b/acceptance/cmd_runner.go
@@ -25,7 +25,7 @@ func NewCmdRunner(logger boshlog.Logger) CmdRunner {
 }
 
 func (r *cmdRunner) RunCommand(env map[string]string, args ...string) (string, string, int, error) {
-	exports := make([]string, len(env))
+	var exports []string
 	for k, v := range env {
 		exports = append(exports, fmt.Sprintf("%s=%s", k, v))
 	}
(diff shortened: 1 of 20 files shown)


xtreme-nitin-ravindran (approved):

deployment/instance/instance.go
```diff
@@ -216,7 +218,7 @@ func (i *instance) Stop(
 	stage biui.Stage,
 ) error {
 
-	return i.shutdown(pingTimeout, pingDelay, skipDrain, true, stage) //; err != nil {
+	return i.shutdown(pingTimeout, pingDelay, skipDrain, true, stage) // ; err != nil {
```
selzoc, line 221: Seems like we should just take out the comment?

acceptance/cmd_runner.go
```diff
@@ -39,7 +39,7 @@ func (r *cmdRunner) RunCommand(env map[string]string, args ...string) (string, s
 }
 
 func (r *cmdRunner) RunStreamingCommand(out io.Writer, env map[string]string, args ...string) (string, string, int, error) {
-	exports := make([]string, len(env))
+	var exports []string
```
selzoc, line 42: What's wrong with `make`ing a slice if you know the length beforehand?

selzoc (approved):
LGTM, some questions just about the results of the linters.

acceptance/cmd_runner.go
rajathagasthya: Nothing wrong, but you'd want to index the slice to add items instead of `append` like below. `append` is going to grow the slice and then you have `len(env)` zero-values at the beginning of the slice.

PR #685, opened 2025-03-28
Director FactoryConfig support for InsecureSkipVerify

kinjelom:
Added `InsecureSkipVerify` support to `Director.FactoryConfig` to allow skipping SSL certificate verification. This is needed for testing a program that connects to a BOSH Director through an SSH tunnel using `bosh_cli` and `director.Factory`.

### DIFF
--- a/director/factory.go
+++ b/director/factory.go
@@ -41,18 +41,23 @@ func (f Factory) New(factoryConfig FactoryConfig, taskReporter TaskReporter, fil
 }
 
 func (f Factory) httpClient(factoryConfig FactoryConfig, taskReporter TaskReporter, fileReporter FileReporter) (Client, error) {
-	certPool, err := factoryConfig.CACertPool()
-	if err != nil {
-		return Client{}, err
-	}
-
-	if certPool == nil {
-		f.logger.Debug(f.logTag, "Using default root CAs")
+	var rawClient *http.Client
+	if factoryConfig.InsecureSkipVerify {
+		f.logger.Debug(f.logTag, "Using insecure SSL connection (skipping server certificate verification)")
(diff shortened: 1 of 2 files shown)


rkoster:
@kinjelom could you elaborate a bit more on your use case? At least for the CLI we have always taken a stance that a valid CA cert is required.

kinjelom:
Sure, that approach makes sense, but isn’t it ultimately up to the user to decide the level of security they are comfortable with? For example, a user can access the REST API via `curl` while ignoring server authentication.

In my case, this option is useful for debugging a connection to a development BOSH director through an SSH tunnel — the HTTPS client sees a different IP address, so the certificate check fails.

klakin-pivotal:
Speaking just for myself (rather than in my official capacity as someone who works on BOSH), I've always been frustrated by the fact that the CLI doesn't just treat the certificate presented by the Director as a very large pre-shared key. It seems to me that in the common case, an operator already tells the CLI what certificate to expect; that the Director presents the certificate you're expecting *should* be proof enough that you're talking to the system you expect to be talking to.

kinjelom:
The change concerns the use of `director.factory` and `uaa.factory` as a library in another program, and this does not affect the functionality of the `bosh-cli` command line.

BTW, have you considered separating the bosh-cli library code base from the application that exposes this functionality via the command line?

beyhan:
ping @rajathagasthya, @xtreme-nitin-ravindran it will be great if you could review this and also the [last comment](https://github.com/cloudfoundry/bosh-cli/pull/685#issuecomment-2778151456) there is a question.

aramprice:
@kinjelom - To answer the CLI / Lib split question, this would be a nice change. There are folks who consume (parts of) the bosh-cli code as a library and having this be a cleaner separation would be great.

PR #687, opened 2025-04-09
Don't package match when given a tarball

rkoster:
Package matching is great for itterating on a release, but when a tarball is provided, the changes of amatch are slim to none. Before this change, tarballs would be repackaged even if no matches where found.

### DIFF
--- a/cmd/upload_release.go
+++ b/cmd/upload_release.go
@@ -90,7 +90,7 @@ func (c UploadReleaseCmd) uploadFile(opts UploadReleaseOpts) error {
 		return bosherr.Errorf("Cannot upload non-remote release")
 	}
 
-	releaseReader, releaseDir := c.releaseDirFactory(opts.Directory)
+	_, releaseDir := c.releaseDirFactory(opts.Directory)
 
 	var release boshrel.Release
 	var err error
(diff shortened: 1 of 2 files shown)


cmd/upload_release.go
```diff
@@ -90,18 +90,15 @@ func (c UploadReleaseCmd) uploadFile(opts UploadReleaseOpts) error {
...
 	var release boshrel.Release
 	var err error
 
 	path := opts.Args.URL.FilePath()
 
 	if len(path) > 0 {
-		release, err = releaseReader.Read(path)
-		if err != nil {
-			return err
-		}
```
klakin-pivotal, line 104: This looks like some sort of sanity check for the local release we're about to upload. Why is this being removed?

(Sadly, the check was committed in a huge commit a decade ago whose description is 'cli', so we don't have a detailed explanation for why it exists.)

rkoster: I believe this check is to see if a tar is being passed (as a file) the tar is then loaded into a release object. 
My change is getting rid of the intermediate step (when a file is being passed) and just directly go to upload.

jpalermo:
Concern we found at the FIWG meeting. It seems likely that the current change breaks uploading a release yaml file.

PR #696, opened 2025-09-22
Add --no-compression flag to bosh create-release command

mdzhigarov:
This PR adds a --no-compression flag to the bosh create-release command, allowing users to create releases without compressing archives.

## Changes:
- Add NoCompression option to CreateReleaseOpts
- Update create-release command to accept --no-compression flag  
- Modify archive writer to support no-compression mode
- Update job renderer and package compiler to pass compression option
- Add compression option to stemcell operations
- Update bosh-utils dependency to use custom fork with compression options
- Add tests for no-compression functionality

## Benefits:
- Useful for debugging release creation issues
- Faster processing when working with large files
- Maintains backward compatibility (compression is still the default)

## Testing:
- All existing tests pass
- New tests added for no-compression functionality
- Binary builds successfully with the new flag

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -170,7 +170,7 @@ func (c Cmd) Execute() (cmdErr error) {
 			return releaseReader, releaseDir
 		}
 
-		releaseWriter := relProv.NewArchiveWriter()
+		releaseWriter := relProv.NewArchiveWriter(false)
 
 		releaseArchiveFactory := func(path string) boshdir.ReleaseArchive {
 			return boshdir.NewFSReleaseArchive(path, deps.FS)
@@ -446,7 +446,7 @@ func (c Cmd) Execute() (cmdErr error) {
 
 		_, err := NewCreateReleaseCmd(
 			releaseDirFactory,
-			relProv.NewArchiveWriter(),
+			relProv.NewArchiveWriter(opts.NoCompression),
 			c.deps.FS,
 			c.deps.UI,
 		).Run(*opts)
@@ -457,7 +457,7 @@ func (c Cmd) Execute() (cmdErr error) {
 
 		return NewRedigestReleaseCmd(
 			relProv.NewArchiveReader(),
-			relProv.NewArchiveWriter(),
+			relProv.NewArchiveWriter(false),
 			crypto.NewDigestCalculator(c.deps.FS, []boshcrypto.Algorithm{boshcrypto.DigestAlgorithmSHA1}),
 			boshfu.NewFileMover(c.deps.FS),
 			c.deps.FS,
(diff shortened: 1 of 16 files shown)


cmd/cmd.go
```diff
@@ -446,7 +446,9 @@ func (c Cmd) Execute() (cmdErr error) {
 
 		_, err := NewCreateReleaseCmd(
 			releaseDirFactory,
-			relProv.NewArchiveWriter(),
+			// TODO: expose noCompression option to create release cmd once all changes are merged and promoted
```
rkoster: what are "all changes" in this context?

mdzhigarov: There are PRs opened for bosh-utils, bosh-agent and bosh (director) that are ready for review. Once they are merged, I have to update this final piece and test the whole solution end-to-end

mdzhigarov:
> In general I don't like the ux of passing a flag, It would be better to be able to disable compression through the package spec file. This gives more fine-grained control and a deterministic result (always same result in depended of flags being passed).

That does mean we'd need to change the packaging spec format (add new field) but also the release. MF manifest schema so that we have info about whether the whole tar should be compressed or not. 

Do you agree? I would like that approach even better. 

Is there a forum where I can discuss this stuff live?

mdzhigarov:
I updated the PR. Now the implementation is much cleaner - the `bosh create-release` takes a --no-compression flag (currently commented out) which is used to control the compression of the final tar.
Each Package spec contains a no_compression: true/false property that can be used to control the compression of the package tars individually. 

I tested this extensively with a mix of compressed and non-compressed options and it seems to produce the proper POSIX tars.

The PR is ready for review and potentially merging.

@rkoster would you please check this today? I need to get this merged so that I can update the Ops Manager dependency to bosh cli. Currently OpsMan uses `bosh deploy` and `bosh upload-release` to upload the releases onto the Director so it will benefit from the no_compression flag in the manifest because apparently to bosh cli re-archives the tar during upload-release/deploy.

rkoster (changes requested):

cmd/create_release.go
```diff
@@ -99,7 +99,7 @@ func (c CreateReleaseCmd) buildRelease(releaseDir boshreldir.ReleaseDir, opts Cr
 		}
 	}
 
-	return releaseDir.BuildRelease(name, version, opts.Force)
```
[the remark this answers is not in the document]
mdzhigarov: Good point! This is not needed anymore. Will clean up the signature.

cmd/create_release.go
```diff
@@ -99,7 +99,19 @@ func (c CreateReleaseCmd) buildRelease(releaseDir boshreldir.ReleaseDir, opts Cr
...
+	release, err := releaseDir.BuildRelease(name, version, opts.Force)
+	if err != nil {
+		return nil, err
+	}
+
+	// Check for no_compression mismatch and warn if found
+	hasMismatch, packages, err := releaseDir.CheckNoCompressionMismatch()
+	if err == nil && hasMismatch {
+		packageList := strings.Join(packages, ", ")
+		c.ui.ErrorLinef("Warning: The following packages have no_compression: true in their spec files, but final.yml does not have no_compression: true. Consider setting no_compression: true in final.yml to avoid double compression: %s", packageList)
```
rkoster: Compression on the outer release tarball

rkoster (changes requested):

rkoster (approved):

beyhan (approved):
LGTM. @mdzhigarov Thank you for the prompt reaction on the discussions and reviews.

PR #740, opened 2026-09-22
Manually bump go-humanize to 1.1.0

ystros:
go-humanize 1.1.0 changes the rounding behavior of humanize.Bytes. Previously, 123456 KB was rounded up to 124 MB. Now, it is rounded properly to 123 MB.

Since we assert the output text, this manual bump is necessary to get past the test failure.


KauzClay (approved):

KauzClay:
should fix https://bosh.ci.cloudfoundry.org/teams/main/pipelines/bosh-cli/jobs/bump-deps/builds/165

